PR #4, opened 2015-02-25
Add tasks to create/remove dev access rules

allait:
Creates AWS security group rules to enable remote access to Elasticsearch and RDS database from given IPs

I haven't moved SSH access out of the Elasticsearch CloudFormation template since it's required for provisioning (and so needs to be in the same state as elasticsearch tasks). 

SSH access for all Elastic Beanstalk instances is always enabled by Amazon.

### DIFF
--- a/playbooks/roles/aws_digitalmarketplace_api/tasks/main.yml
+++ b/playbooks/roles/aws_digitalmarketplace_api/tasks/main.yml
@@ -33,3 +33,18 @@
   register: digitalmarketplace_api_env_stack
   tags:
     - digitalmarketplace-api
+
+- name: RDS dev access security group rules ({{dev_access_state}})
+  cloudformation:
+    stack_name: "digitalmarketplace-api-{{ name_suffix }}-dev-access"
+    state: "{{ dev_access_state }}"
+    region: "{{ aws_region }}"
+    disable_rollback: false
+    template: ../cloudformation_templates/aws_digitalmarketplace_api_rds_dev_access.json
+    template_parameters:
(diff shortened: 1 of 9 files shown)


playbooks/roles/aws_digitalmarketplace_api/tasks/main.yml
```diff
@@ -33,3 +33,18 @@
...
+  cloudformation:
+    stack_name: "digitalmarketplace-api-{{ name_suffix }}-dev-access"
+    state: "{{ dev_access_state }}"
+    region: "{{ aws_region }}"
+    disable_rollback: false
+    template: ../cloudformation_templates/aws_digitalmarketplace_api_rds_dev_access.json
+    template_parameters:
+      CidrIp: "{{ user_ip }}"
+      RDSSecurityGroup: "{{ digitalmarketplace_api_env_stack.stack_outputs.RDSSecurityGroup }}"
+  register: digitalmarketplace_api_rds_dev_access_stack
```
robyoung: These registered variables get really long. We should think about a terse convention for them.

robyoung:
Other than the big long registered variables this looks good.

playbooks/roles/aws_digitalmarketplace_api/tasks/main.yml
allait: I'm trying to use a simple consistent naming scheme for now, so staying away from abbreviations.  This way the vars that end up in common/user var files hopefully will be self-descriptive. But I do agree that a terse convention might be better if we come up with a good one.

PR #5, opened 2015-02-27
Add user.yml config sample

allait:


robyoung:
I'm not entirely sure of the purpose of this. Could we get a bit more explanation for this please? Ideally a bit more on the 'why' in the commit message as well.

See https://github.com/alphagov/styleguides/blob/master/git.md#commit-messages and https://github.com/alphagov/styleguides/blob/master/pull-requests.md for the GOV.UK guidance on commits and pull requests which is pretty good.

PR #6, opened 2015-02-27
Add IAM Instance Profile to digitalmarketplace-api Elastic Beanstalk instances

allait:
List of permissions from http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/AWSHowTo.iam.roles.aeb.html

> The following example shows one statement that gives a broad set of permissions to AWS products that AWS Elastic Beanstalk uses to manage applications and environments and includes permissions to create an instance profile and view a list of available instance profiles

### DIFF
--- a/cloudformation_templates/aws_digitalmarketplace_api_env.json
+++ b/cloudformation_templates/aws_digitalmarketplace_api_env.json
@@ -79,6 +79,48 @@
       }
     },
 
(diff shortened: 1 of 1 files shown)


robyoung:
Could we get a bit more of an explanation of why these particular IAM permissions are being added in the commit message? Even if it's just a link off to the AWS docs.

PR #9, opened 2015-03-03
Add Elasticsearch ELB and ELB URLs to stack outputs

allait:
Creates an ELB in front of Elasticsearch and adds ELB URL to elasticsearch and digitalmarketplace-api stack outputs so that we can use them to configure apps that require access to elasticsearch/API.

### DIFF
--- a/cloudformation_templates/aws_dev_access.json
+++ b/cloudformation_templates/aws_dev_access.json
@@ -9,7 +9,7 @@
     },
     "CidrIp": {
       "Type": "String",
-      "Description": "CIDR or IP to use for security group rules"
+      "Description": "CIDR IP to use for security group rules"
     },
     "FromPort": {
       "Type": "Number",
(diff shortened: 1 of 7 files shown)


cloudformation_templates/aws_dev_access.json
```diff
@@ -9,7 +9,7 @@
     },
     "CidrIp": {
       "Type": "String",
-      "Description": "CIDR or IP to use for security group rules"
```
robyoung: I don't think the 'or' should have been removed here.

TheDoubleK: Looking at the docs it seems the or should go?
"CidrIp - Specifies a CIDR range."
 http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html#cfn-ec2-security-group-ingress-cidrip

PR #11, opened 2015-03-05
Dm auth tokens

TheDoubleK:


allait:
Will merge this once it's updated with the new env variable name and without the `.gitignore` change

PR #13, opened 2015-03-16
A new CLI based on stacks.yml

allait:
New create/delete/provision/deploy commands interface:

    dmaws <stage> <environment> <command> [apps] [command options and arguments]

`create`/`delete` commands use `stacks.yml` for CloudFormation stacks instead of Ansible.
`deploy` command creates an archive with repository contents and build artefacts


dmaws/build.py
```diff
@@ -0,0 +1,82 @@
...
+
+    run_git_cmd(['archive', '--format=zip', 'HEAD'], cwd, stdout=package_file)
+
+    os.close(package_file)
+
+    return ref, sha, file_path
+
+
+def run_project_build_script(cwd):
+    run_cmd(['./scripts/build.sh'], cwd=cwd)
```
TheDoubleK: Where is this `build.sh` script? Can't see it in this project or in the apps being deployed?

allait: It's run from the repository being deployed. I have a script for the admin-frontend that I'll send for review a bit later.

TheDoubleK:
It would be good to have the instructions documented in the README as well as the pull request message, just to keep it current.

allait:
@TheDoubleK there's `--help` for the CLI that has a full list of commands and accepted arguments/options for each one. I'll look into adding it to README once we're happy with the CLI.

TheDoubleK:
OK fair enough.

dmaws/commands/provision.py
```diff
@@ -0,0 +1,34 @@
...
+import json
+import subprocess
+
+from ..cli import cli_command
+from ..utils import run_cmd
+
+
+def run_playbook(playbook, hosts, ctx, basedir='playbooks/'):
+    args = [
+        'ansible-playbook',
```
TheDoubleK: So (just to clarify) do we still need Ansible for 'provision' but are dropping it for create/delete/deploy?

vars/production.yml
```diff
@@ -0,0 +1,5 @@
+---
+subnets:
+  - subnet-bf8c5ce6
+  - subnet-916bfcf4
+  - subnet-82dc63f5
```
TheDoubleK: The `production.yml` and `preview.yml` subnets are identical - is that right? And if so do we need both files, or could we just have one file that is used for both stages?

dmaws/commands/provision.py
allait: Yeah. And we're gathering the variables in python and sending the full dictionary to ansible, so it's only responsible for actually running the provisioning tasks

PR #17, opened 2015-03-19
Add instance type parameters

allait:
Requires #16


vars/common.yml
```diff
@@ -2,11 +2,26 @@
...
+
+search_api:
+  instance_type: t2.micro
+  min_instance_count: 1
+  max_instance_count: 2
+
+admin_frontend:
+  instance_type: t2.micro
+  min_instance_count: 1
+  max_instance_count: 2
```
robyoung: Do these need to be overridden int he `production.yml`?

vars/common.yml
```diff
@@ -2,11 +2,26 @@
...
+
+database:
+  port: 5432
+  user: "digitalmarketplace"
+  name: "digitalmarketplace_api"
+  instance_type: db.t2.micro
+  allocated_storage: 5
 
 elasticsearch:
   port: 9200
```
robyoung: Should the elasticsearch cluster be configurable in the same way?

allait: It already is but it just has `instance_count` instead of `min_` and `max_` because we can't autoscale it

vars/common.yml
allait: Yes, but I think we should do this once we have the production account and know the final instance sizes that we'll use

PR #19, opened 2015-03-23
Clone app when releasing

robyoung:
The release command is going to be called from Jenkins. This means that
there will not be an existing, safe chekout of the app repositories.
This commit solves that problem by changing release to accept an app
name and then cloning the app's repo locally.

### DIFF
--- a/dmaws/build.py
+++ b/dmaws/build.py
@@ -12,6 +12,8 @@ HTTPS_REPO_PATTERN = re.compile('https://[^/]+/[^/]+/(.*)/(?:.git)?')
 RELEASE_TAG_PATTERN = re.compile(r'^release-\d+$')
 MERGE_COMMIT_PATTERN = re.compile(r'[a-f0-9]+ Merge pull request #(\d+) from')
 
+REPOS_PATH = '.repos'
+
 
 def run_git_cmd(args, cwd, stdout=None):
     return run_cmd(
@@ -19,6 +21,25 @@ def run_git_cmd(args, cwd, stdout=None):
         cwd=cwd,
         stdout=stdout or subprocess.PIPE
     )
+
+
+def clone_or_update(app_name):
+    app_git_url = get_application_git_url(app_name)
+    repository_path = os.path.join(REPOS_PATH,
+                                   'digitalmarketplace-{}'.format(app_name))
+    if not os.path.exists(REPOS_PATH):
+        os.mkdir(REPOS_PATH)
+    if not os.path.exists(repository_path):
+        run_git_cmd(['clone', app_git_url], REPOS_PATH)
+    else:
+        run_git_cmd(['reset', '--hard', 'origin'], repository_path)
+        run_git_cmd(['pull'], repository_path)
+
+    return repository_path
+
+
+def get_application_git_url(app_name):
+    return 'git@github.com:alphagov/digitalmarketplace-{}.git'.format(app_name)
 
 
 def get_application_name(cwd):
@@ -79,7 +100,8 @@ def get_release_name_for_tag(cwd, tag):
     if len(tags) == 1:
         return tags[0]
     elif len(tags) > 1:
-        raise StandardError("Something strange is going on, we have more than one release tag pointing to same commit")
+        raise StandardError(
+            "More than one release tag pointing to the same commit.")
 
 
 def get_release_name_for_repo(cwd):
--- a/dmaws/build.py
+++ b/dmaws/build.py
@@ -32,8 +32,11 @@ def clone_or_update(repo_url):
     if not os.path.exists(repository_path):
         run_git_cmd(['clone', repo_url], REPOS_PATH)
     else:
-        run_git_cmd(['reset', '--hard', 'origin'], repository_path)
-        run_git_cmd(['pull'], repository_path)
+        run_git_cmd(['reset', '--hard', 'HEAD'], repository_path)
+        run_git_cmd(['checkout', 'master'], repository_path)
+        run_git_cmd(['clean', '-fdx'], repository_path)
+        run_git_cmd(['fetch'], repository_path)
+        run_git_cmd(['reset', '--hard', 'origin/master'], repository_path)
 
     return repository_path
 
--- a/dmaws/build.py
+++ b/dmaws/build.py
@@ -23,14 +23,14 @@ def run_git_cmd(args, cwd, stdout=None):
     )
 
 
-def clone_or_update(app_name):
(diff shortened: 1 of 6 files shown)


dmaws/build.py
```diff
@@ -21,6 +23,25 @@ def run_git_cmd(args, cwd, stdout=None):
...
+        run_git_cmd(['clone', app_git_url], REPOS_PATH)
+    else:
+        run_git_cmd(['reset', '--hard', 'origin'], repository_path)
+        run_git_cmd(['pull'], repository_path)
+
+    return repository_path
+
+
+def get_application_git_url(app_name):
+    return 'git@github.com:alphagov/digitalmarketplace-{}.git'.format(app_name)
```
allait: Can we use the repository URL as the command argument instead? That way we only rely on the `digitalmarketplace-{app_name}` convention and can deploy from other clones if needed

robyoung: I would be reluctant to allow us to release from other clones unless we have a good reason. I think it's fine for the development account but for production I'd want it more explicit.

dmaws/build.py
```diff
@@ -21,8 +23,22 @@ def run_git_cmd(args, cwd, stdout=None):
...
+def clone_or_update(repo_url):
+    app_name = get_application_name_from_url(repo_url)
+    repository_path = os.path.join(REPOS_PATH,
+                                   'digitalmarketplace-{}'.format(app_name))
+    if not os.path.exists(REPOS_PATH):
+        os.mkdir(REPOS_PATH)
+    if not os.path.exists(repository_path):
+        run_git_cmd(['clone', repo_url], REPOS_PATH)
+    else:
+        run_git_cmd(['reset', '--hard', 'origin'], repository_path)
```
allait: We should remove all untracked and ignored files at this point too, including build artefacts from previous builds and npm/bower modules

robyoung: Good spot, this is in fact completely wrong, `git reset --hard origin` isn't even a valid command.

scripts/jenkins-release.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+[ -d venv ] || virtualenv venv
+pip install -e .
+
+if [ "$ENVIRONMENT" = "Select one" -o "$ENVIRONMENT" = "" ]; then
```
allait: This is probably a stage more than environment (even though it's used as both)

scripts/jenkins-release.sh
```diff
@@ -0,0 +1,26 @@
...
+  >&2 echo "You must select an environment."
+  exit 1
+fi
+
+if [ "$APPLICATION_NAME" = "Select one" -o "$ENVIRONMENT" = ""]; then
+  >&2 echo "You must select an application name."
+  exit 1
+fi
+
+if [ "$ENVIRONMENT" = "preview" ]; then
```
allait: What do you think about having one AWS_PROFILE per stage, instead of one per AWS account (i.e. preview/staging/production profiles)? That way we could either use the same access keys for staging/production and get the same behaviour or look into separating permissions where possible

dmaws/build.py
```diff
@@ -79,7 +101,8 @@ def get_release_name_for_tag(cwd, tag):
     if len(tags) == 1:
         return tags[0]
     elif len(tags) > 1:
-        raise StandardError("Something strange is going on, we have more than one release tag pointing to same commit")
+        raise StandardError(
```
allait: ValueError?

dmaws/build.py
```diff
@@ -21,8 +23,24 @@ def run_git_cmd(args, cwd, stdout=None):
...
+                                   'digitalmarketplace-{}'.format(app_name))
+    if not os.path.exists(REPOS_PATH):
+        os.mkdir(REPOS_PATH)
+    if not os.path.exists(repository_path):
+        run_git_cmd(['clone', repo_url], REPOS_PATH)
+    else:
+        run_git_cmd(['reset', '--hard', 'HEAD'], repository_path)
+        run_git_cmd(['checkout', 'master'], repository_path)
+        run_git_cmd(['clean', '-fdx'], repository_path)
+        run_git_cmd(['pull'], repository_path)
```
allait: This is probably safe enough for us, but one case where this won't work as expected is if origin master was force-pushed. How about `fetch` followed by `git reset --hard origin/master`?

scripts/jenkins-release.sh
robyoung: I went with environment because it's a more commonly understood term. It's the language that's used throughout GDS and the service manual.

dmaws/build.py
robyoung: Yea, ok...

scripts/jenkins-release.sh
robyoung: At some point we have to have the knowledge of which stages are associated with which accounts. I'd rather that be recorded somewhere in git, the jenkins credentials file will probably not be.

scripts/jenkins-release.sh
allait: That's a good point. I'm not sure if Jenkins is the place we need to be worried about being commonly understood since it's less accessible than the deployment repository itself. At the same time, I feel like it's best to be consistent in our usage of `stage` vs `environment`, otherwise it can get confusing.

Maybe it's worth changing `stage` to `environment` and `environment` to `??` since this keeps coming up.

scripts/jenkins-release.sh
robyoung: After chatting in real life we've agreed to document this in the manual.

dmaws/build.py
robyoung: Yep, good spot.

scripts/jenkins-release.sh
robyoung: Agreed in real life to rename this variable to STAGE and then look at renaming STAGE to ENVIRONMENT and ENVIRONMENT to ENVIRONMENT_CONTEXT.

scripts/jenkins-release.sh
```diff
@@ -0,0 +1,20 @@
...
+  exit 1
+fi
+
+if [ "$APPLICATION_NAME" = "Select one" -o "$STAGE" = ""]; then
+  >&2 echo "You must select an application name."
+  exit 1
+fi
+
+if [ "$STAGE" = "staging" ]; then
+  OPTIONS="--from-account=development --release-name=$RELEASE_NAME"
```
allait: `--from-account=preview`?

allait:
:+1:

PR #26, opened 2015-03-25
Upgrade instances to Amazon Linux 2014.09 v1.2.0

ajlanghorn:
Upgrade Digital Marketplace instances running in Elastic Beanstalk to Amazon Linux 2014.09 v1.2.0.

### DIFF
--- a/cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
+++ b/cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
@@ -92,7 +92,7 @@
       "Properties": {
         "ApplicationName": {"Ref": "ApplicationName"},
         "Description": "Default Configuration Version 1.0",
-        "SolutionStackName": "64bit Amazon Linux 2014.09 v1.1.0 running Python 2.7",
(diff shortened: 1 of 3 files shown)


robyoung:
How have these been tested?

ajlanghorn:
They haven't been yet, @robyoung. That was the reason these haven't yet been merged.

I don't know how you do your testing, other than I suppose bringing up new Beanstalk instances.

robyoung:
I think that would be it. We have some smoke tests but they haven't been wired up yet. I would think that i we can deploy an app to the new instance type and the smoke tests pass then it's probably going to be ok.

TheDoubleK:
As soon as we have URLs for the deployed apps we can easily point the smoke tests at them to run - configurable in the Jenkins job.

allait:
I've tested this in preview and it seems to be working with admin-frontend and api.
We had problems with deployment when moving from 1.0.9 to 1.1.0 and it appears to be ok with 1.2.0, so I'm :+1: to merge.

ajlanghorn:
\o/

PR #27, opened 2015-03-26
Run CloudFormation templates through Jinja

allait:
Takes all `EnvVar*` parameters from the stack parameters list and:

1. Adds a parameter declaration to CloudFormation template
2. Adds OptionSetting to set parameter value to environment variable with the `param_to_env(name)` in the Elastic Beanstalk environment

Loaded CloudFormation template is run through jinja2 using selected values from the stack as context.

Right now the only variables available to jinja are:

* `environment_variables` – a subset of parameter names and
* `param_to_env` – a transform function to get environment variable
name out of parameter name.

Parameter is added to 'environment_variables' if its name starts with 'EnvVar' (CloudFormation doesn't allow underscores in parameter names so direct mapping using actual environment variable names doesn't work. And 'Env' appears in 'EnvironmentName').

This is a first step to make changing CloudFormation templates easier and less error-prone and reducing duplication. Jinja will allow us to use includes and macros for commonly repeated parts of templates as well as generating list of accepted parameters from stacks info.

I'm not moving the environment variables to a new stack field separate from parameters yet, since it's a bigger change, but it's something we might do at some point.

### DIFF
--- a/dmaws/stacks.py
+++ b/dmaws/stacks.py
@@ -1,3 +1,5 @@
+import re
+
 from toposort import toposort_flatten
 
 from .cloudformation import Cloudformation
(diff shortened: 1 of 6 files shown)


dmaws/stacks.py
```diff
@@ -34,14 +36,39 @@ def __init__(self, stack, stage, environment, variables):
...
+        template_body = load_file(self.template)
+
+        environment_variables = [
+            p for p in self.parameters.keys() if p.startswith('EnvVar')
+        ]
+
+        def param_to_env(name):
+            s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
+            s2 = re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1)
+            return s2.upper().replace('ENV_VAR_', '')
```
robyoung: This took me a bit of thinking to understand. If what we're trying to say is "Prefix any capital letter that is not at the start of the string with an underscore" could this just be done with `re.sub('(.)([A-Z])', r'\1_\2', name)`

dmaws/utils.py
```diff
@@ -79,6 +79,14 @@ def __init__(self, mapping, variables, **kwargs):
         self._cache = {}
         self._variables = merge_dicts(variables, kwargs)
 
+    def keys(self):
+        return self._mapping.keys()
+
+    def items(self):
+        for key in self.keys():
+            self.__getitem__(key)
+        return self._cache.items()
```
robyoung: Bums, I've implemented this in one of my PRs. Race to first merge!!

dmaws/stacks.py
allait: This splits things like `HTTPRequest` into `http_request` too, while the simpler methods don't. We don't actually _need_ this, since you can use `HttpRequest` in parameter name and it will be uppercased anyway, but I'd still keep the current version.

robyoung: Ahh, I see. Sounds good then.

PR #28, opened 2015-03-26
Wait for ready status on release

robyoung:
Wait for the status of the Beanstalk application to go back to 'Ready'  before exiting. This status unfortunately doesn't tell us whether the release was successful, so also track the event stream, print them and cause dmaws to fail if any errors are encountered.

There are also a couple of little bug fixes to the release process in here as well because here is where I found them. I can move them into a separate pull request if people think it's necessary.

### DIFF
--- a/dmaws/utils.py
+++ b/dmaws/utils.py
@@ -83,9 +83,7 @@ class LazyTemplateMapping(object):
         return self._mapping.keys()
 
     def items(self):
-        for key in self.keys():
-            self.__getitem__(key)
-        return self._cache.items()
+        return [(key, self[key]) for key in self.keys()]
 
     def __getitem__(self, key):
         if key not in self._cache:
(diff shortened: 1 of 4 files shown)


dmaws/utils.py
```diff
@@ -79,6 +79,9 @@ def __init__(self, mapping, variables, **kwargs):
         self._cache = {}
         self._variables = merge_dicts(variables, kwargs)
 
+    def items(self):
```
allait: This will be a merge conflict with https://github.com/alphagov/digitalmarketplace-aws/pull/27/files#diff-82cedc58a55843ad9169fc5b6ea91af6R82 if both land, so we should probably chose whichever looks better and keep it

robyoung: I like a mix of the two actually. I like having an explicit `keys` method and I like the list comprehension based `items` method that knows nothing about `_cache`. Looking at the pull requests as a whole #27 is less contentious and should probably therefore be merged first. Would you object if I changed `items` to the list comprehension style after it's merged?

allait: Sounds good. I'm :+1: with the PR otherwise

PR #31, opened 2015-03-27
[WIP] Add cloudfront

robyoung:
THIS IS FOR REVIEW ONLY - PLEASE DO NOT MERGE YET

This introduces a basic and not-yet-ready-for-prime-time stack for
configuring the API in CloudFront. This is not ready for prime time
because CloudFront talks to Beanstalk over HTTP. In order to have have
CloudFront talk to Beanstalk over SSL we need to serve Beanstalk over a
domain name that we can generate an SSL certificate for. We can do this
without managing DNS in Route53 but it means the only way we're wiring
the CloudFront configuration up to a Beanstalk instance is through DNS
that is managed elsewhere which seems non-obvious.

Other things of note:
- The S3 template has been renamed to be more generic.
- This introduces a new output to the API template.
- The domain names follow the convention covered in the Google Doc shared
  within the team.
- All logs for a stage are stored in the same bucket with different
  prefixes.
- The IAM certificate is generated separately and must be provided.
- The root domain is configurable to allow the different accounts to be
  on different root domains as discussed.

### DIFF
--- a/cloudformation_templates/aws_cloudfront_app.json
+++ b/cloudformation_templates/aws_cloudfront_app.json
@@ -1,0 +1,72 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Description": "Digital Marketplace Application CloudFront Distribution",
+
+  "Parameters": {
+    "OriginDomain": {
+      "Type": "String",
+      "Description": "The origin domain for the upstream application."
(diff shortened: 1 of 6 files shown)


robyoung:
Closing in favour of getting the Route53 changes merged separately first.

PR #37, opened 2015-03-31
Add Route53 managed domains for all apps

robyoung:
Create a Route53 zone and then create domains for each app of the form
`{ stage }-{ environment }-{ app name }.{ root domain }`

This is a precursor to setting up SSL and CloudFront.

### DIFF
--- a/cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
+++ b/cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
@@ -33,6 +33,14 @@
     "EnvironmentName": {
       "Type": "String",
       "Description": "Elastic Beanstalk Environment name"
+    },
+    "Domain": {
+      "Type": "String",
+      "Description": "Domain to serve the Beanstalk environment from"
+    },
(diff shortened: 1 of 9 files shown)


cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
```diff
@@ -156,6 +186,10 @@
       "Description": "Digital Marketplace EB environment",
       "Value": {"Ref": "Environment"}
     },
+    "Subdomain": {
+      "Description": "The subdomain of the app",
+      "Value": {"Ref": "Subdomain"}
+    },
     "URL": {
```
allait: We should return the URL with the new domain here, since it's used to link the apps together. We'll just have to switch it to https when the certificates is in place.

allait:
This looks good, but I think we need to set up the `[development.]digitalmarketplace.` DNS before merging this, otherwise links between apps will stop working if we update any environment with this.

So let's run the `route53zone` stack locally, set up the DNS and then merge and update.

PR #38, opened 2015-04-01
Add loaddata command and some related changes

allait:
Downloads a database dump from S3 and runs it on the environment RDS instance.

Runs Elasticsearch import script from the digitalmarketplace-api repository with environment API and Elasticsearch as arguments.

A lot of this will change once we have write endpoints on api and search-api that would allow loading data throught them instead of using the databases directly.

### DIFF
--- a/dmaws/commands/loaddata.py
+++ b/dmaws/commands/loaddata.py
@@ -1,0 +1,79 @@
+import os
+import sys
+
+import click
+import boto.ec2
+import requests
+
+from ..cli import cli_command
+from ..stacks import StackPlan
+from ..utils import run_cmd
+
+from ..deploy import S3Client
+
+LOADDATA_BUCKET = 'digitalmarketplace-dev-loaddata'
+
+
+@cli_command('loaddata', max_apps=0)
+@click.argument('api_repo_path')
+@click.argument('tasks', nargs=-1)
+def loaddata_cmd(ctx, api_repo_path, tasks):
+    """Load test data for development or preview environment."""
+
+    if not tasks:
+        tasks = ['database', 'search']
+
+    plan = StackPlan.from_ctx(ctx, apps=['dev_access'])
+
+    status = plan.create(create_dependencies=False)
+    if not status:
+        sys.exit(1)
+
+    plan.info(['api'])
+
+    elasticsearch_ip = get_tagged_ec2_instances(
+        ctx.variables['aws_region'],
+        'elasticsearch-{}-{}'.format(ctx.stage, ctx.environment)
+    )[0].ip_address
+
+    db_path = "postgres://{}:{}@{}".format(
+        plan.get_value('database.user'),
+        plan.get_value('database.password'),
+        plan.get_value('stacks.database.outputs.URL')
+    )
+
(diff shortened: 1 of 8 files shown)


robyoung:
Some of the commit messages could benefit from a bit of explanation to why they're being done. For example, using the `KeyName` type; was it causing a specific problem or was this just to use a more specific type? [git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md)

allait:
@robyoung no idea how it's actually different from String tbh, I saw there was a separate type for keys, so I changed the templates. I've read the style guide before.

robyoung:
That in itself is useful information.

dmaws/commands/loaddata.py
```diff
@@ -0,0 +1,79 @@
...
+        es_endpoint = 'http://{}:9200/services/'.format(elasticsearch_ip)
+        requests.put(es_endpoint)
+
+        ctx.log('Inserting data into Elasticsearch')
+        run_cmd([
+            './scripts/process-g6-into-elastic-search.py',
+            es_endpoint,
+            plan.get_value('stacks.api.outputs.URL') + '/services',
+            plan.get_value('api.auth_tokens')[0]
+        ], cwd=api_repo_path, ignore_errors=True)
```
robyoung: Why do we ignore errors here?

allait: There's no step that depends on this being successful, and we still need to remove downloaded file and do the dev-access teardown

dmaws/commands/loaddata.py
```diff
@@ -0,0 +1,79 @@
...
+        ], ignore_errors=True)
+
+    if 'search' in tasks:
+        ctx.log('Creating Elasticsearch index')
+        es_endpoint = 'http://{}:9200/services/'.format(elasticsearch_ip)
+        requests.put(es_endpoint)
+
+        ctx.log('Inserting data into Elasticsearch')
+        run_cmd([
+            './scripts/process-g6-into-elastic-search.py',
```
robyoung: Where is this script from?

allait: API repo path, which we take as an argument. It'll get replaced with something more suitable as soon as the API is pushing updates to Elasticsearch

dmaws/commands/loaddata.py
robyoung: I think it would still be useful to know that it has failed.

allait: The command output is printed in any case

dmaws/commands/loaddata.py
robyoung: Ah, I see, missed the `cwd`.

dmaws/commands/loaddata.py
robyoung: It swallows the fact that there was any error. If the command being run doesn't output anything you have no feedback that it failed. A log message to say that the command has failed would be helpful.

allait: Added start and finish log messages to run_cmd. Prints the command name only and return code on exit.

allait:
Rebased and fixed merge conflict on KeyName type change

PR #39, opened 2015-04-02
Common base jinja template for Elastic Beanstalk apps

allait:
Base template defines common parameters, resources and outputs, including EnvVar pattern and Elastic Beanstalk options.

Per-app templates should extend app_base.j2 and can define:
* additional parameters in {% block parameters %}
* additional resources in {% block resources %}
* additional outputs in {% block outputs %}
* override Elastic Beanstalk solution stack in {%block solution_stack %}


robyoung:
Why does the ElasticBeanstalk solution stack need to be overridden with Jinja block?

allait:
@robyoung so that we're not forced to update all apps at once

PR #46, opened 2015-04-08
Update the EB solution stack to 2015.03 v1.3.0

allait:
Release notes for the 2015.03 Amazon Linux AMI:
http://aws.amazon.com/amazon-linux-ami/2015.03-release-notes/

Applied to all preview apps.

### DIFF
--- a/cloudformation_templates/app_base.j2
+++ b/cloudformation_templates/app_base.j2
@@ -105,7 +105,7 @@
       "Properties": {
         "ApplicationName": {"Ref": "ApplicationName"},
         "Description": "Elastic Beanstalk Configuration",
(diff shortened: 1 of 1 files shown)


robyoung:
Do we know this will be applied in a 'no downtime' way?

robyoung:
Python 3.4 is available...

allait:
@robyoung it should be the same as any other environment update, but we don't have any live environments yet anyway.

I think Python 3 is something we can actually consider since we're doing a pretty good job at maintaining compatibility (mostly thanks to @quis, who runs the admin app on py3 locally). But I'd wait to see if any of our future dependencies might be an issue. And it feels like a larger switch that requires more testing than the AMI update.

robyoung:
About Py3, yea, totally.

PR #51, opened 2015-04-13
Add the supplier and admin frontends to cloudfront

robyoung:
This changes the byers app to be the default origin. It also serves the supplier app under `/supplier` and the admin app under `/admin`. During the architecture meeting we seemed to all agree that the admin app should live on a subdomain rather than through CloudFront. What was the consensus on that (cc/ @minglis).

This has some implications for the supplier and admin apps. All their routes must be under `/supplier` and `/admin` respectively or they will not work through the www sub-domain. This will require some changes.


quis:
> byers app

Deliberate typo?

robyoung:
:D

PR #57, opened 2015-04-16
Set static URL prefixes for admin and supplier apps

allait:
http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options.html
docs mention that the /static/ option name can be any arbitrary value.

Uses template block since there's no parameter that has the correct
value and the admin/supplier paths are hardcoded in the CloudFront template.

### DIFF
--- a/cloudformation_templates/app_base.j2
+++ b/cloudformation_templates/app_base.j2
@@ -146,7 +146,7 @@
 {% endfor %}
           {
             "Namespace": "aws:elasticbeanstalk:container:python:staticfiles",
-            "OptionName": "/static/",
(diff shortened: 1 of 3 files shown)


robyoung:
I don't understand the reasoning for not having these configured in the same way that all other parameters are.

allait:
@robyoung they're not parametrised in the `www_cloudfront` template. So they're not dynamic and just show the difference between the frontend templates. We've used parameters for this kind of configuration before (e.g. for dev access and S3 buckets), but it seems that the template inheritance is a better fit here.

PR #58, opened 2015-04-16
CloudWatch logs

robyoung:
This change gives the beanstalk EC2 instances access to write log events to CloudWatch and also creates the CloudWatch log group to be used by all apps in the environment.


robyoung:
oops, one sec, only tested on api

PR #61, opened 2015-04-17
Add flag to ignore dependencies on delete command

robyoung:


dmaws/stacks.py
```diff
@@ -156,7 +156,7 @@ def create(self, create_dependencies=True):
 
         return True
 
-    def delete(self):
+    def delete(self, ignore_deps=False):
```
allait: `deps` could be both dependencies and dependants, it might be better to have the full word here, just like we do in `create(self, create_dependencies=True)`

allait:
It would be nice to add a test for `StackPlan.delete`

PR #68, opened 2015-04-24
AMI based nginx and Elasticsearch instances

allait:
Replaces provisioning with custom-build AMIs and user-data based cloud-init configuration.

AMIs are created using packer templates and stored in the AWS account (currently on development account).
CloudFormation then sets up the user-data to a script that would kick-off the instance-specific configuration on instance launch.


cloudformation_templates/aws_nginx.json
```diff
@@ -0,0 +1,283 @@
...
+            "#!/bin/bash","\n",
+            "cd /home/ubuntu/provisioning && ",
+            "ansible-playbook -c local -i localhost, nginx_playbook.yml ",
+            "-t instance-config ",
+            "-e assets_domain=", {"Ref": "AssetsDomain"}, " ",
+            "-e documents_s3_url=", {"Ref": "DocumentsS3URL"}, " ",
+            "-e www_domain=", {"Ref": "Domain"}, " ",
+            "-e buyer_frontend_url=", {"Ref": "BuyerFrontendURL"}, " ",
+            "-e admin_frontend_url=", {"Ref": "AdminFrontendURL"}, " ",
+            "-e supplier_frontend_url=", {"Ref": "SupplierFrontendURL"}
```
robyoung: Surprisingly nice.

robyoung:
Could you add a bit to the README? Particularly running packer.

allait:
@robyoung the `/404` redirect will serve the buyer-frontend 404 page instead of the nginx or S3 default ones. The main reason for this are the `403` responses for missing S3 documents. There's no handler for the 5xx response, so the S3 one will be served. But I'm not sure what they look like and what we should do in this case.

PR #70, opened 2015-04-29
Update API env vars for dmutils

robyoung:
dmutils standardises the environment variables holding the API URLs and
auth tokens. This change leaves the old environment variables in place
to avoid breaking changes during the switch over.

### DIFF
--- a/stacks.yml
+++ b/stacks.yml
@@ -82,6 +82,7 @@ api:
   template: cloudformation_templates/aws_digitalmarketplace_api.json
   dependencies:
     - api_app
+    - search_api
     - database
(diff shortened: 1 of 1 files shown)


stacks.yml
```diff
@@ -299,6 +307,8 @@ supplier_frontend:
     EnvVarDmEnvironment: "{{ stage }}"
     EnvVarDmApiUrl: "{{ stacks.api.outputs.URL }}"
```
allait: Are the `EnvVarDmApiUrl` and `ApiAuthToken` still used then?

PR #71, opened 2015-04-30
nginx and elasticsearch AMI cloudwatch logs

allait:
Adds cloudwatch streams for nginx and elasticsearch.
Provided awslogs configs will send nginx access_log and general elasticsearch cluster log to cloudwatch.
New AMIs contiain the cloudwatch ansible role and have awslogs preinstalled.

### DIFF
--- a/playbooks/roles/cloudwatch/templates/elasticsearch_awslogs.conf.j2
+++ b/playbooks/roles/cloudwatch/templates/elasticsearch_awslogs.conf.j2
@@ -1,0 +1,8 @@
+[general]
+state_file = /var/awslogs/state/agent-state
+
+[elasticsearch-log]
+file = /var/log/elasticsearch/{{ elasticsearch_name }}.log
+log_group_name = {{ cloudwatch_log_group }}
+log_stream_name = {{ elasticsearch_name }}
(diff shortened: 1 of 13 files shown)


playbooks/roles/cloudwatch/templates/elasticsearch_awslogs.conf.j2
```diff
@@ -0,0 +1,8 @@
+[general]
+state_file = /var/awslogs/state/agent-state
+
+[elasticsearch-log]
+file = /var/log/elasticsearch/{{ elasticsearch_name }}.log
+log_group_name = {{ cloudwatch_log_group }}
+log_stream_name = {{ elasticsearch_name }}
```
robyoung: Why does elasticsearch use the `GroupTag` but Nginx doesn't?

allait: I'm not sure if that's what you mean, but GroupTag is used for elasticsearch cluster discovery (so that nodes only join within their own environment). Since nginx doesn't cluster there's no need for GroupTag.

The stream name is a bit redundant, `elasticsearch-preview-preview` inside the `preview-preview` group. But since it's also the cluster name, I think it makes sense.

PR #72, opened 2015-05-05
Create and delete exit status

allait:
This is a lot of small stuff bundled together, I can split them into separate pull requests if needed.

* Returns stack create/delete/update errors as exit code 1 (I'll silence the return code in Jenkins for now, since it will always fail due to EB apps being in DELETE_FAILED state)
* Cleans up and structures log output
* Renames the elasticsearch ELB to meet the length requirements for production (requires recreating the ASG, so the cluster will be wiped)
* Fixes an error with Cloudformation RateLimit exceptions being treated as "stack missing" responses
* Restores python3 compatibility since ansible is no longer a requirement

### DIFF
--- a/stacks.yml
+++ b/stacks.yml
@@ -1,5 +1,6 @@
 ---
 all:
+  - www
   - apps
   - dev_access
   - data_storage
@@ -30,6 +31,7 @@ data_storage:
   - documents_s3
   - logs_s3
   - elasticsearch
+  - monitoring
 
 dev_access:
   - database_dev_access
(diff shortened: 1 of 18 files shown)


stacks.yml
```diff
@@ -30,10 +31,13 @@ data_storage:
   - documents_s3
   - logs_s3
   - elasticsearch
+  - monitoring
```
robyoung: I don't think this should be here. The specific stacks that require it should depend on it.

allait: Delete traverses dependencies in the opposite order. If monitoring isn't listed in the `all` group the stack won't get deleted unless explicitly mentioned

allait:
Sneaked in more indentation/color changes for the deployment output

PR #73, opened 2015-05-06
Add new environment vars for sending password reset emails

TheDoubleK:
To send emails the app needs a Mandrill key, secret password salt and a 'namespacing' salt.
This adds these three parameters as environment variables for the supplier app.

This PR should be merged before: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/17

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -25,6 +25,7 @@ supplier_frontend:
   instance_type: t2.micro
   min_instance_count: 1
   max_instance_count: 2
+  password_salt: ResetPasswordSalt
 
 database:
   port: 5432
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -25,7 +25,6 @@ supplier_frontend:
   instance_type: t2.micro
   min_instance_count: 1
   max_instance_count: 2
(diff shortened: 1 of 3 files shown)


vars/common.yml
```diff
@@ -25,6 +25,7 @@ supplier_frontend:
   instance_type: t2.micro
   min_instance_count: 1
   max_instance_count: 2
+  password_salt: ResetPasswordSalt
```
allait: I'd keep the password_salt in the secret config, just to be on the safe side. Otherwise there's no reason to keep it as environment variable, it can just be set in app config directly

TheDoubleK:
I've added secret things to the Jenkins .yml files.

I've used a test Mandrill key for preview, but we'll need to bear in mind that staging will send emails for real (unless we want to split the staging and prod secret config file).

vars/common.yml
TheDoubleK: OK maybe I'll do that - i.e. move it into app config file. From what it said in the docs you linked to I don't think it needs to be secret.

PR #75, opened 2015-05-06
Add MultiAZ parameter to RDS instances

robyoung:
In development and preview RDS instances do not need to be MultiAZ in
staging and production they do.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-multiaz


vars/development.yml
```diff
@@ -2,6 +2,7 @@
 root_domain: "development.digitalmarketplace.service.gov.uk"
 ssl_certificate_id: "arn:aws:iam::381494870249:server-certificate/cloudfront/star-development-digitalmarketplace/star-development-digitalmarketplace"
 iam_certificate_id: "ASCAIJITQ6F5D5P4NWJZS"
+rds_multi_az: true
```
allait: PR description says "In development and preview RDS instances do not need to be MultiAZ", but this is `true`.

robyoung: That is a very good point :no_mouth:.

PR #77, opened 2015-05-08
Set CloudWatch metric namespace on apps

robyoung:


TheDoubleK:
Empty pull-request message :(

PR #78, opened 2015-05-08
Internal EB environments

allait:
Makes EB environment ELBs internal to the VPC.

* ELB now listens for HTTP connections, since all traffic is internal to the VPC
* Private Route53 records for the EB apps
* Closed SSH access to EB instances by default
* Updated nginx config with glob server_names and api config


stacks.yml
```diff
@@ -356,12 +351,15 @@ nginx:
...
     AdminFrontendURL: "{{stacks.admin_frontend.outputs.URL}}"
     BuyerFrontendURL: "{{stacks.buyer_frontend.outputs.URL}}"
     SupplierFrontendURL: "{{stacks.supplier_frontend.outputs.URL}}"
 
-    Domain: "{% if stage != 'production' %}{{ stage }}{% else %}www{% endif %}{% if stage != environment %}-{{ environment }}{% endif %}.{{ root_domain }}"
-    AssetsDomain: "{% if stage != 'production' %}{{ stage }}-{% endif %}{% if stage != environment %}{{ environment }}-{% endif %}assets.{{ root_domain }}"
+    RootDomain: "{{ root_domain }}"
+    Subdomain: "{% if stage != 'production' %}{{ stage }}{% else %}www{% endif %}{% if stage != environment %}-{{ environment }}{% endif %}"
+    ApiSubdomain: "{% if stage != 'production' %}{{ stage }}-{% endif %}{% if stage != environment %}{{ environment }}-{% endif %}api"
+    AssetsSubdomain: "{% if stage != 'production' %}{{ stage }}-{% endif %}{% if stage != environment %}{{ environment }}-{% endif %}assets"
```
robyoung: I really like that all these conditions are in one place now.

PR #80, opened 2015-05-12
Configure nginx to re-resolve upstream hosts

allait:
When proxy_pass is used with a predefined URL nginx will resolve
the host during startup and will cache the IP address indefinitely.

This is a problem when proxying requests to AWS ELBs, since ELB IPs
change without notice.

To force nginx to re-resolve DNS proxy_pass destination must be set
to a variable and resolver must be defined in server configuration.

Nginx can't use the nameserver IP from system resolv.conf, so we're
passing it in as another variable.

Since we set resolver in nginx.conf and the variable is only set
in cloud-init stage nginx.conf is now run during instance-config.

nameserver_ip is set by user data cloud-init script by running the `resolv.conf` through `awk`.

### DIFF
--- a/cloudformation_templates/aws_nginx.json
+++ b/cloudformation_templates/aws_nginx.json
@@ -294,6 +294,7 @@
             "ansible-playbook -c local -i localhost, nginx_playbook.yml ",
             "-t instance-config ",
             "-e aws_region=", {"Ref": "AWS::Region"}, " ",
+            "-e nameserver_ip=$(awk '/nameserver/{ print $2; exit}' /etc/resolv.conf)",
             "-e cloudwatch_log_group=", {"Ref": "LogGroupName"}, " ",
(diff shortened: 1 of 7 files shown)


TheDoubleK:
All looks good as far as I can tell. One question: why does the tag need to change from `image-setup` to `instance-config`?

allait:
@TheDoubleK `image-setup` is run by packer during AMI creation whereas `instance-config` is started by cloud-init user-data script when instance is launched. Since the resolver is in the `nginx.conf` creating the config requires the `nameserver_ip` variable to exist, which is only true for the cloud-init stage. (Also, I'm not sure if resolv.conf is the same on the live instance and during AMI creation)

We could set resolvers in each individual `server { ... }` blocks instead of once in `http { ... }`, but at the same time I can think of no real benefit to having nginx.conf be part of the AMI.

PR #81, opened 2015-05-13
[#92327956] Add RDS backup related fields to CF

robyoung:
Add RDS backup related fields to the CloudFormation template.

## BackupRetentionPeriod
Number of days to keep automatic snapshots around. Set to 1 for
non-production and 30 for production.
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-BackupRetentionPeriod

## DBSnapshotIdentifier
A snapshot identifier to use to restore a database from a snapshot. It
is defaulted to blank. The idea being it would be provided as a custom
variable to restore an environment to a snapshot.
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbsnapshotidentifier

### DIFF
--- a/cloudformation_templates/aws_rds_database.json
+++ b/cloudformation_templates/aws_rds_database.json
@@ -28,6 +28,14 @@
     "MultiAZ": {
       "Type": "String",
       "Description": "Whether the instance is multi AZ or not"
+    },
+    "BackupRetentionPeriod": {
+      "Type": "String",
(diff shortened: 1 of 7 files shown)


robyoung:
Closing while I investigate CloudFormation and RDS a bit more.

PR #83, opened 2015-05-13
Add rds backup and snapshot

robyoung:
### BackupRetentionPeriod
Number of days to keep automatic snapshots around. Set to 1 for
non-production and 30 for production.
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-BackupRetentionPeriod

### DBSnapshotIdentifier
A snapshot identifier to use to restore a database from a snapshot. It is defaulted to blank. The idea being it would be provided as a custom variable to restore an environment to a snapshot.
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbsnapshotidentifier

## 	Add support for RDS snapshots

A few things get added in this commit:
- A DeletionPolicy is added to the DBInstance resource to Snapshot when
  the resource is deleted
- If DBSnapshotIdentifier is provided DBName is not provided. If a
  DBName is provided a new empty database is always created so the two
  options cannot be provided together.

This approach is largely taken from this blog post
http://blog.jasonantman.com/2014/12/aws-cloudformation-and-rds-snapshots/

## 	Restore preview to a snapshot

This snapshot was made after loading all suppliers, services and users
from production.

## Update from testing

- Deleting the database stack causes an email to be sent
- Updating the database stack requiring replacement causes an email to be sent

### DIFF
--- a/cloudformation_templates/aws_monitoring.json
+++ b/cloudformation_templates/aws_monitoring.json
@@ -20,6 +20,29 @@
         "LogGroupName": {"Ref": "LogGroupName"},
         "RetentionInDays": {"Ref": "RetentionInDays"}
       }
+    },
+
+    "RDSSNSTopic": {
+      "Type": "AWS::SNS::Topic",
+      "Properties": {
+        "DisplayName": "RDSEvents",
+        "Subscription": [
+          {
+            "Protocol": "email",
+            "Endpoint": "support@digitalmarketplace.service.gov.uk"
+          }
+        ]
+      }
+    },
+
+    "RDSEventSubscription": {
+      "Type": "AWS::RDS::EventSubscription",
+      "Properties": {
+        "Enabled": true,
+        "EventCategories": ["deletion"],
+        "SnsTopicArn": {"Ref": "RDSSNSTopic"},
+        "SourceType": "db-instance"
+      }
     }
   },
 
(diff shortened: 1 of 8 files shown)


cloudformation_templates/aws_monitoring.json
```diff
@@ -20,6 +20,28 @@
...
+    },
+
+    "RDSSNSTopic": {
+      "Type": "AWS::SNS::Topic",
+      "Properties": {
+        "DisplayName": "RDSEvents",
+        "Subscription": [
+          {
+            "Protocol": "email",
+            "Endpoint": "support@digitalmarketplace.service.gov.uk"
```
allait: Let's move this to a variable in case we'll have different mailing lists for environments in the future

robyoung: Good call. Done.

vars/development.yml
```diff
@@ -2,7 +2,10 @@
 root_domain: "development.digitalmarketplace.service.gov.uk"
 ssl_certificate_id: "arn:aws:iam::381494870249:server-certificate/cloudfront/star-development-digitalmarketplace/star-development-digitalmarketplace"
 iam_certificate_id: "ASCAIJITQ6F5D5P4NWJZS"
-rds_multi_az: "false"
+
+rds:
```
allait: Let's merge `rds` with `database` variable

PR #97, opened 2015-06-10
Update frontend artifacts paths included in the release

allait:
Expects project build script to output artefact paths to be included
into the release archive. One path per line.

This requires the build script to redirect any other output to STDERR.
No longer includes any paths on top of the git archive by default.

Adding directory to archive verifies that path is relative to the
project root to prevent root path traversal (eg doesn't allow adding
'/' or '../')

### DIFF
--- a/dmaws/build.py
+++ b/dmaws/build.py
@@ -122,8 +122,9 @@ def get_release_name_for_repo(cwd):
 
 
 def add_directory_to_archive(cwd, path, archive_path):
+    full_path = utils.safe_path_join(cwd, path)
     with zipfile.ZipFile(archive_path, 'a') as archive:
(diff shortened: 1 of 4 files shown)


quis:
I think these are the pull requests that require merging in order for this to be merged:
- https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/57
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/75
- https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/67

quis:
Looks good to go as far as I'm concerned, once the others are merged.

quis:
Scratch that; this will break the loading of the content in the admin app. Where would be a good place to copy the files to? `app/content` maybe?

PR #98, opened 2015-06-10
Update nginx to 1.8

allait:
Uses nginx apt repository to install the latest version
Allows installing nginx 1.8.0 on Ubuntu 14.04

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -43,4 +43,4 @@ elasticsearch:
 nginx:
   instance_type: t2.micro
   instance_count: 2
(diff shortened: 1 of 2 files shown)


vars/common.yml
```diff
@@ -43,4 +43,4 @@ elasticsearch:
 nginx:
   instance_type: t2.micro
   instance_count: 2
-  instance_image: ami-2b99ee5c
+  instance_image: ami-d94739ae
```
TheDoubleK: Out of interest, where does this ami number come from?

PR #106, opened 2015-06-26
Add S3 bucket for G7 draft documents

robyoung:
Documents for draft services will go into a separate bucket that is not
visible to the outside world. After submissions close we will virus
scan them before migrating only the successfull ones into the live
docuemnts S3 bucket.


TheDoubleK:
Does this also need changes for the environment variables that were added/removed?

robyoung:
Good call! I'll do that now.

PR #122, opened 2015-07-22
Add a CloudWatch log group for JSON logs

robyoung:
Performing more complex metric filters is much easier with JSON logs.

See: http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/FilterAndPatternSyntax.html#d0e11216


pcraig3:
:+1:

PR #123, opened 2015-07-22
Fix deploy command docstring

robyoung:
If this blank line is in the docstring then click blows up. This is a bug in click but it's easy to fix here.


allait:
:+1:

PR #124, opened 2015-07-23
Add JSONLogGroupName to the application stacks

robyoung:
It is easier and less error prone to pass this in to the application stacks rather than generating it in each application stack.


allait:
:+1:

PR #125, opened 2015-07-30
nginx robots.txt, max request size and ASG update policy

allait:
### Set nginx request size limit to 40Mb to allow file uploads
Default `client_max_body_size` is 1Mb and nginx returns 413 if
requests exceeds the size limit.

### Add robots.txt to block /admin and /suppliers indexing
[#99500206](https://www.pivotaltracker.com/story/show/99500206)

Adds robots.txt file served from the root of www. Disallows any
URLs starting with `/admin` or `/suppliers`, so that both root
pages and all other pages of admin and supplier apps are not
indexed.

Since suppliers-guide was moved to `/g-cloud/suppliers-guide`,
there's no need to add an allow rule for it.

### Set nginx ASG health check to use ELB status
By default Auto Scaling Groups use EC2 instance status
checks to add or remove instances from the group. Which
means that even if the instance doesn't respond to the
ELB health checks it will still be "InService".

Changing the ASG health check to "ELB" makes sure that only
instances that pass all attached ELB health checks are added
to the group and any instances that are reported as "OutOfService"
by the load balancer are replaced.

The health check grace period set to 5 minutes ensures that instance
has enough time to boot up before the health check is performed,
otherwise launching instances could be marked as unhealthy before
that have finished setting up nginx.

(http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/as-add-elb-healthcheck.html)

### Add cloudformation policy to cycle nginx instances on update
Setting update policy on the AutoScalingGroup resource makes
cloudformation replace instances if the launch configuration
has changed. This way there's no need to terminate the nginx
instances manually after AMI changes.

With ELB-based health checks set for the nginx ASG setting
batch size and min number of live instances should be enough
to make sure that Cloudformation replaces instances one at a
time and waits for the instance to complete setup steps before
shutting down the next one.

(http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html)

### DIFF
--- a/cloudformation_templates/aws_nginx.json
+++ b/cloudformation_templates/aws_nginx.json
@@ -293,6 +293,12 @@
         "MinSize" : {"Ref": "InstanceCount"},
         "MaxSize" : {"Ref": "InstanceCount"},
         "DesiredCapacity" : {"Ref": "InstanceCount"}
+      },
+      "UpdatePolicy" : {
+        "AutoScalingRollingUpdate": {
+          "MinInstancesInService": "1",
+          "MaxBatchSize": "1"
+        }
       }
     },
 
--- a/cloudformation_templates/aws_nginx.json
+++ b/cloudformation_templates/aws_nginx.json
@@ -288,6 +288,8 @@
         "AvailabilityZones" : {"Fn::GetAZs": ""},
         "LaunchConfigurationName" : {"Ref": "LaunchConfiguration"},
         "LoadBalancerNames": [{"Ref": "LoadBalancer"}],
+        "HealthCheckType": "ELB",
+        "HealthCheckGracePeriod": 300,
         "MinSize" : {"Ref": "InstanceCount"},
         "MaxSize" : {"Ref": "InstanceCount"},
         "DesiredCapacity" : {"Ref": "InstanceCount"}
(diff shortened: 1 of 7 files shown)


cloudformation_templates/aws_nginx.json
```diff
@@ -288,9 +288,17 @@
         "AvailabilityZones" : {"Fn::GetAZs": ""},
         "LaunchConfigurationName" : {"Ref": "LaunchConfiguration"},
         "LoadBalancerNames": [{"Ref": "LoadBalancer"}],
+        "HealthCheckType": "ELB",
+        "HealthCheckGracePeriod": 300,
```
robyoung: 5 minutes sounds like quite a long grace period. Does this affect how quickly it will apply the update?

robyoung:
:+1: Looks good. Is it worth adding a story to do a similar thing for the elasticsearch cluster? It would need to be a bit more sophisticated, checking the cluster health.

cloudformation_templates/aws_nginx.json
allait: Yes, since it should update instances one by one this will act as an additional delay for each instance.

I was trying to figure out a bit more about this as it's not clear when the timer starts (is it the instance start time or the time between EC2 health check and ELB one, since default EC2 health check seemed to work fine without it etc.). But most things I found recommend setting it to a high value, mostly because otherwise ASG can fail to bring up new instances and just restart them over and over again.

robyoung: Fair point. :shipit:

PR #126, opened 2015-07-30
nginx connection draining

allait:
### Enable connection draining for nginx ELB
Connection draining will keep instance InService for the given
time to allow existing requests to finish. This way nginx AMI
updates could be applied without interrupting or failing any
requests.

This is the same setting we've enabled to fix failing requests on
application deploys.

### Add /static forwarding to buyer frontend for assets domain
404 page forwarded from the buyer frontend to handle missing service
documents references static files required to display the page.

When handling requests for /static URLs nginx tries to request them
from S3, gets a 404 and returns the same buyer frontend 404 page.

Adding a /static location forwarded to the buyer frontend fixes the
issue.

### DIFF
--- a/cloudformation_templates/aws_nginx.json
+++ b/cloudformation_templates/aws_nginx.json
@@ -200,6 +200,10 @@
         "SecurityGroups": [
           {"Fn::GetAtt": ["LoadBalancerSecurityGroup", "GroupId"]}
         ],
+        "ConnectionDrainingPolicy": {
+          "Enabled": true,
+          "Timeout": 300
+        },
         "HealthCheck": {
           "Target": "HTTP:80/_status",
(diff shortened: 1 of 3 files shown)


cloudformation_templates/aws_nginx.json
```diff
@@ -200,6 +200,10 @@
         "SecurityGroups": [
           {"Fn::GetAtt": ["LoadBalancerSecurityGroup", "GroupId"]}
         ],
+        "ConnectionDraining": {
+          "Enabled": true,
+          "Timeout": 300
```
minglis: 300 whats? is that millis or seconds? Is it long?

minglis:
👍

PR #128, opened 2015-08-06
User IPs ingress for www ELB

allait:
## Add UserIPs parameter to nginx stack
[#99310600](https://www.pivotaltracker.com/story/show/99310600)

Sets the ELB ingress rule to only allow connections from the given
IPs/CIDR ranges. Defaults to 0.0.0.0/0 in vars/common.

Overwriting the `user_ips` list in user.yml allows limiting access
to full preview/staging environments.

## Use a list of dev_user_ips for ssh/dev-access tasks
Creates a separate ingress resource for each developer IP/CIDR range.

Requires:
- [ ]  digitalmarketplace-credentials variables update (PR 21)

### DIFF
--- a/cloudformation_templates/aws_dev_access.json
+++ b/cloudformation_templates/aws_dev_access.json
@@ -7,9 +7,9 @@
       "Type": "String",
       "Description": "Target security group name"
     },
-    "CidrIp": {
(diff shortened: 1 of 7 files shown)


robyoung:
:+1:

PR #129, opened 2015-08-07
Proxy g-cloud-7 assets to the draft documents bucket

allait:
Maps assets /g-cloud-7 prefix to the G7 draft documents bucket.
This makes it possible to use the assets domain for draft documents
as well as any framework related files (eg supplier pack) as long as
the S3 object path starts with "g-cloud-7" prefix.

### DIFF
--- a/cloudformation_templates/aws_nginx.json
+++ b/cloudformation_templates/aws_nginx.json
@@ -308,6 +308,7 @@
       },
       "UpdatePolicy" : {
         "AutoScalingRollingUpdate": {
+          "PauseTime" : "PT10M",
           "MinInstancesInService": "1",
           "MaxBatchSize": "1"
         }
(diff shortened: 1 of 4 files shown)


robyoung:
:+1:

allait:
Updated nginx AMI.

### Set nginx ASG update pause
allait authored 10 minutes ago
ELB-based health checks for ASG instances are not used for adding
the instance to the pool, only for removing the current instances.
Instances are added to the pool as soon as the EC2 check finishes,
and health check grace period only affects the ELB health checks
that would remove the InService instance.

So the ELB health check doesn't add the necessary delay required
for nginx instances to finish setup when replacing instances during
the launch configuration update.

Adding pause time forces Cloudformation to wait between relaunching
instances. This would work better if we had `"WaitOnResourceSignals"`
set up, which would continue the update once the instance reported
as ready, but for now the 10 minute delay should be enough for nginx
to come up without increasing the update time too much.

PR #130, opened 2015-08-07
Add JSON stream for nginx log

allait:
Adds nginx logs to the json stream group using a log format similar
to the one used by Apache logs.


robyoung:
:+1:

PR #132, opened 2015-08-21
Add request times to nginx logs

robyoung:
Add $request_time to access logs.
http://nginx.org/en/docs/http/ngx_http_log_module.html


allait:
:+1: 

(I usually test the nginx config changes before merging by SSHing into one of the preview machines, replacing the file in-place and reloading. It's hard to spot typos in these things otherwise and rolling back an AMI update is a bit arduous.)

PR #134, opened 2015-08-27
Add a CloudWatch metric for Nginx requestTime

robyoung:
This will create a CloudWatch metric filter that takes the requestTime from Nginx JSON logs and puts it into a CloudWatch metric. In preview the metric will have a namespace preview-preview/Nginx and a name of requestTime. This doesn't match metric naming convention we've used to date because we don't have access to the ElasticBeanstalk environment name. This may actually turn out to be a more flexible naming convention.

This has been tested on preview

screen shot 2015-08-27 at 16 01 34


allait:
Is there a way to set the seconds as units?
Looks good otherwise :+1:

robyoung:
It is seconds.

PR #135, opened 2015-08-27
DM-Request-ID updates for Nginx logs

robyoung:
## Use upstream DM-Request-ID for Nginx logs

$http_dm_request_id takes the value from the request header however we
want to take it from the response of the upstream.

http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_http_

## Unset DM-Request-ID on initial requests

The way that we pass the DM-Request-ID around services is through the HTTP header. As a result, a client could set the DM-Request-ID themselves from outside. This strips that header on the way in so that one is generated internally.

There is no unset header directive, instead an empty value prevents the header being passed on to the proxied server.

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header

### DIFF
--- a/playbooks/roles/nginx/templates/api.j2
+++ b/playbooks/roles/nginx/templates/api.j2
@@ -10,6 +10,7 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto https;
         proxy_set_header Host $http_host;
+        proxy_set_header DM-Request-ID "";
 
         proxy_redirect http:// https://;
 
@@ -29,6 +30,7 @@ server {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto https;
         proxy_set_header Host $http_host;
+        proxy_set_header DM-Request-ID "";
 
         proxy_redirect http:// https://;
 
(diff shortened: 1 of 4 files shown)


TheDoubleK:
It's an old forum thread, but assuming they haven't changed anything in the meantime this looks like it should work :+1: 

BTW, Where does the AMI number come from? (I'm sure @allait explained it to me once before but maybe I need to be told over again until I remember).

robyoung:
When you run the [`packer`](https://github.com/alphagov/digitalmarketplace-aws#creating-amis-with-packer) command the AMI number is outputted at the end.

robyoung:
Also, that's a good point about the old forum thread, I'm going to try and find it in the docs and reference that instead.

allait:
We don't set the request header on the responses ATM, so this still requires a change to the apps.
And should we consider stripping the request/response headers between nginx and the user? I don't think they're sensitive, but I'd rather not leave them modifiable from the outside.

robyoung:
@allait actually we do https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/logging.py#L30

allait:
@robyoung cool, I somehow missed that. Stripping incoming headers sounds good.

allait:
`proxy_set_header` is valid within http and server context so we can set it in the `nginx.conf` once or in each server instead of each location

PR #137, opened 2015-09-03
Move 5xx metric filters and alarms to dmaws

robyoung:
Move the 5xx metric filters and alarms into dmaws from the applications .ebextensions.

A short version of the application name has been added to the DM_METRICS_NAMESPACE to match the namespace used for Nginx logs and be easier to search for than making the metric names themselves really long. This is the AWS recommended way of organising metrics.

Metric filters and alarms have been pulled out into macros to avoid mistakes in boilerplate such as log group name, alarm actions and namespace.

Alarm descriptions are implemented in separate templates so that they can be easily managed and more easily made uniform (eg. extending from a base template).

After discussion we have decided not to port over the 4xx and related alarms unless they can be shown to be useful. Instead favouring more specific alarms where needed.

### DIFF
--- a/cloudformation_templates/alarms/CloudWatchHTTP5xxCountAlarm.j2
+++ b/cloudformation_templates/alarms/CloudWatchHTTP5xxCountAlarm.j2
@@ -1,0 +1,14 @@
+{
+  "Fn::Join": [
+    "",
+    [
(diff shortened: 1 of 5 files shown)


robyoung:
Closing while I fix a bug

PR #138, opened 2015-09-04
Move cloudwatch to dmaws

robyoung:
This has been tested on preview and can be seen in the CloudWatch console. Select the `preview-preview/api` metric in the Custom Metrics drop down on the left.

## Move 5xx metric filters and alarms to dmaws
Move the 5xx metric filters and alarms into dmaws from the applications .ebextensions.

A short version of the application name has been added to the `DM_METRICS_NAMESPACE` to match the namespace used for Nginx logs and be easier to search for than making the metric names themselves really long. This is the AWS recommended way of organising metrics. For example, the metrics namespace for the supplier app in preview would be `preview-preview/supplier-frontend`.

Metric filters and alarms have been pulled out into macros to avoid mistakes in boilerplate such as log group name, alarm actions and namespace.

The metric filter macro accepts a sub-part of the patter which it wraps and adds the application filter to automatically. For example, giving the metric filter `$.foo = bar` to the macro will result in a metric
filter of `{ $.application = api && ($.foo = bar) }` on the api app.

Alarm descriptions are implemented in separate templates so that they can be easily managed and more easily made uniform (eg. extending from a base template).

After discussion we have decided not to port over the 4xx and related alarms unless they can be shown to be useful. Instead favouring more specific alarms where needed.

## Add {ACTION}_ROLLBACK_FAILED as complete condition
This was seen when a CloudFormation internal failure happened during a rollback.

## Add apache request time metric filter
Add a CloudWatch metric filter for Apache requestTimeMicro so that we can graph individual application performance.

## Add metrics and alarm for supplier login failure
Adds metric filters for login failure and success and an alarm that fires if more than 20% of login attempts fail.

## 	Add 4xx metric filters
Add a variety of 4xx related metric filters to gain a better idea of what kind of errors we get here. There are no alarms set on these filters as yet as the intention right now is just to get a better idea of what's going on.

### DIFF
--- a/cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
+++ b/cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
@@ -1,5 +1,5 @@
 {% extends "alarms/base_alarm.j2" %}
 
-{% block description %}Login failure rate too high{% endblock %}
+{% block description %}Too many 5xx{% endblock %}
 
-{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#login-failure-rate{% endblock %}
+{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#5xx-errors{% endblock %}
--- a/cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
+++ b/cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
@@ -1,0 +1,5 @@
+{% extends "alarms/base_alarm.j2" %}
+
+{% block description %}Login failure rate too high{% endblock %}
+
+{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#login-failure-rate{% endblock %}
--- a/cloudformation_templates/alarms/LoginFailurePercentAlarm.j2
+++ b/cloudformation_templates/alarms/LoginFailurePercentAlarm.j2
@@ -1,0 +1,5 @@
+{% extends "alarms/base_alarm.j2" %}
+
+{% block description %}Too many 5xx{% endblock %}
+
+{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#5xx-errors{% endblock %}
--- a/cloudformation_templates/alarms/LoginFailurePercentAlarm.j2
+++ b/cloudformation_templates/alarms/LoginFailurePercentAlarm.j2
@@ -1,5 +1,5 @@
 {% extends "alarms/base_alarm.j2" %}
 
-{% block description %}Too many 5xx{% endblock %}
+{% block description %}Login failure rate too high{% endblock %}
 
-{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#5xx-errors{% endblock %}
+{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#login-failure-rate{% endblock %}
--- a/cloudformation_templates/app_base.j2
+++ b/cloudformation_templates/app_base.j2
@@ -297,7 +297,7 @@
                     statistic="Sum",
                     period=60,
                     evaluation_periods=1,
-                    threshold=10,
+                    threshold=1,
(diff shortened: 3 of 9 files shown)


cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
```diff
@@ -0,0 +1,5 @@
+{% extends "alarms/base_alarm.j2" %}
+
+{% block description %}Login failure rate too high{% endblock %}
+
+{% block manual_url %}https://github.gds/pages/gds/digitalmarketplace-manual/alerts.html#login-failure-rate{% endblock %}
```
TheDoubleK: Are we going to get these things written up in the manual at some point? Seems pointless to keep the URLs in otherwise.

cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
```diff
@@ -0,0 +1,5 @@
+{% extends "alarms/base_alarm.j2" %}
+
+{% block description %}Login failure rate too high{% endblock %}
```
TheDoubleK: The file is HTTP5xx alarm but the message is about login failures - doesn't seem right?

cloudformation_templates/alarms/LoginFailurePercentAlarm.j2
```diff
@@ -0,0 +1,5 @@
+{% extends "alarms/base_alarm.j2" %}
+
+{% block description %}Too many 5xx{% endblock %}
```
TheDoubleK: Yep - I think you have the messages round the wrong way - the Login Failure one should be here?

cloudformation_templates/app_base.j2
```diff
@@ -281,6 +282,74 @@
...
+                             pattern="$.logType = apache-access && $.status != 5*",
+                             metric_name="HTTP5xx",
+                             metric_value=0) }},
+
+    {{ metrics.alarm("HTTP5xxCountAlarm",
+                    metric_name="HTTP5xx",
+                    statistic="Sum",
+                    period=60,
+                    evaluation_periods=1,
+                    threshold=10,
```
TheDoubleK: 500s feel like something we could alert on with fewer than ten - in that we shouldn't ever see them and if we do then the app might need to be fixed.  So maybe even just one as a threshold?

cloudformation_templates/aws_digitalmarketplace_supplier_frontend.json
```diff
@@ -1,3 +1,28 @@
 {% extends "app_base.j2" %}
```
TheDoubleK: This doesn't look like a `.json` file - should this also be `.j2`?  Or is Cloudformation happy with this as JSON?

cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
robyoung: Yes, I'm going to do that today.

cloudformation_templates/app_base.j2
robyoung: Agreed, I copied over the original threshold with the idea that I'd reduce them in a separate pull request which wouldn't block this one if there was disagreement. I think two people is enough though.

cloudformation_templates/aws_digitalmarketplace_supplier_frontend.json
robyoung: Our dmaws tool runs this through jinja and the result is valid json.

cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
robyoung: Bums, good spot.

cloudformation_templates/alarms/HTTP5xxCountAlarm.j2
TheDoubleK: Sweet :)

TheDoubleK:
Looks good as far as I can tell - try it out on Preview I reckon.

robyoung:
On preview for all apps and looking good.

TheDoubleK:
:+1:

PR #139, opened 2015-09-07
More metric filter fixes

robyoung:
## Fix 5xx link to manual
Sphinx doesn't support anchors starting with digits for some reason.

## Add admin frontend login failure metric
This is a direct copy of the login failure metric and alarm used for supplier frontend.


TheDoubleK:
Looks good to go. :+1: 

Although it isn't completely obvious (to me at least) when the alarm will fire -  does `threshold=0.20` mean if 20% of login attempts fail during a minute?

robyoung:
20% over the last 5 minutes.

PR #141, opened 2015-09-08
Add a script to find slow routes from CloudWatch logs

robyoung:
This is a pretty hacky script that downloads a week's worth of logs from
CloudWatch and then tries to find the 90th percentile of slowest routes.


TheDoubleK:
As a hacky script for ad-hoc digging into things this seems fine to me :+1:

PR #143, opened 2015-09-10
Change Jenkins Route53 record to A

robyoung:
The ElasticIP is an IP address and therefore the DNS record should be an
A record rather than a CNAME.


pcraig3:
:+1:

PR #144, opened 2015-09-10
Set server tokens to off to hide version

minglis:
server tokens set to off to hide nginx verison.

Required for pen test


TheDoubleK:
:+1:

PR #146, opened 2015-09-18
Update nginx AMI

robyoung:
This is the change committed in
https://github.com/alphagov/digitalmarketplace-aws/pull/144


allait:
:+1:

PR #147, opened 2015-09-18
Update jenkins box

robyoung:
## Make Jenkins instance bigger
We need more memory, the OOM killer keeps killing Jenkins.

## Remove Jenkins instance shutdown behaviour
This prevents the stack running again once you've terminated the instance (because CloudFormation cannot stop it anymore).


allait:
:+1:

With a t2.large we could now enable encryption on the EBS volume if we want, but this would require rebuilding the volume

PR #148, opened 2015-09-23
Allow promoting preview to staging by setting empty release-name

allait:
#### Allow promoting preview to staging by setting empty release-name
If release-name is not set when calling release to staging current
preview release version will be used. This should act similar to the
current production releases.

"release-name" is kept as a parameter and will be used when set. Staging
is the only stage that allows an explicit release name right now, so
this is required for rollbacks (rollbacks should start with staging and
can be promoted to production, preview can not be rolled back right
now).

"release-name" check is changed to `not release_name` to avoid adding
logic to the jenkins script, which will always pass `--release-name=`
for staging, but now an empty value is allowed.

#### Write a KEY=VALUE property file using release command output
Uses ctx.out to write KEY=VALUE pairs to stdout, which is redirected
by the jenkins script to a property file. Stdout is now reserved for
this output format, all other unstructured command output should be
written to stderr by `ctx.log`.

This allows us to communicate release information back to the caller
environment (e.g. Jenkins) to use in the following build steps and
notifications.

### DIFF
--- a/dmaws/commands/release.py
+++ b/dmaws/commands/release.py
@@ -36,6 +36,8 @@ def release_cmd(ctx, release_name=None, from_profile=None):
 
     if not success:
         sys.exit(1)
+
+    ctx.out("RELEASE_NAME", release_name)
     build.push_deployed_to_tag(repository_path, ctx.stage, release_name)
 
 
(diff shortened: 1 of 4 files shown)


quis:
I approve of the intent of this PR

robyoung:
:+1::+1::+1::tada:

PR #151, opened 2015-10-05
Upgrade app EB solution stacks to v2.0.1

allait:
Smoke tests running on the hour are failing with 504 errors. This
turned out to be an issue in the current EB Apache log rotation:

> The root cause of this is that EB rotates the logs once an hour
> BUT when rotating the logs, they do a "reload" on apache rather
> than a 'graceful'. This causes connections to be dropped.

¯\_(ツ)_/¯

and should be fixed in the newer (v1.4.6 and later) stack versions.

(https://forums.aws.amazon.com/thread.jspa?messageID=625900)


robyoung:
As discussed, :+1:, but after G7 close.

PR #152, opened 2015-10-06
Add current production instance sizes to vars file

allait:
Instance sizes for elasticsearch and RDS were set in the credentials
repo. Moving them to aws since they're not supposed to be secret and
keeping them near the default makes it easier to understand what types
of instances we have for different parts of the environment.


robyoung:
:+1:

PR #153, opened 2015-10-08
Fix duplicate database variable in the production vars

allait:
`database` key was created twice, so only the second dictionary
was kept.


TheDoubleK:
What does the `snapshot_id` here do? Seems like a very old one?

TheDoubleK:
But :+1:  for the code fix otherwise.

allait:
@TheDoubleK "snapshot_id" is the only way we found to do DB restores/rollbacks in the future. By setting `snapshot_id` from the start we can change it to a newer snapshot to replace the DB if we need to. We can't remove it since this will trigger a database replacement.

PR #154, opened 2015-10-12
Restore t2.micro instance size for supplier-frontend and api

allait:
Restoring the default instance size for supplier-frontend and Data
API apps after the close of G-Cloud 7 service submissions.

Keeping the autoscaling max instance counts and the medium nginx
instances, since nginx is not autoscaled.


TheDoubleK:
:+1:

PR #156, opened 2015-10-15
Fix 5xx alarm to catch all errors

robyoung:
This alarm was only emailing us if there was more than one 5xx in a 1 minute period.


TheDoubleK:
Amazing how many developers it's taken to spot this one!

:+1:

PR #157, opened 2015-10-16
Add new bucket names

robyoung:
These are the new bucket names agreed in the discussion about URLs

```
communications - framework documents and communications
documents      - publicly visible service documents
submissions    - draft service documents
agreements     - framework agreements
```


TheDoubleK:
:+1:

allait:
New buckets need to be added to dependencies somewhere, otherwise the won't be created as part of the top-level `www` stack.

PR #158, opened 2015-10-19
Add data_storage to www

robyoung:
So that storage that apps depend on gets put in place.

Also removes the `logs_s3` stack as it's no longer needed.


allait:
:+1:

PR #159, opened 2015-10-30
TESTED Add new S3 buckets to Nginx

robyoung:
The agreements, communications and submissions buckets are separated out now. This change makes Nginx proxy pass requests to them.

It also adds the required environment variables to the supplier and admin frontends.

This has been tested on preview.


allait:
:+1:

PR #160, opened 2015-11-06
Continue to use existing G7 draft bucket

robyoung:
We need to preserve the update times for when we list and they cannot be
manually edited.

### DIFF
--- a/playbooks/roles/nginx/templates/assets.j2
+++ b/playbooks/roles/nginx/templates/assets.j2
@@ -26,6 +26,14 @@ server {
         proxy_pass $agreements_s3_url;
     }
 
+    # Hack to force g7 communications to the old bucket to preserve upload times
+    location ^~ /g-cloud-7-updates/communications {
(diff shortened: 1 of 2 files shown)


playbooks/roles/nginx/templates/assets.j2
```diff
@@ -26,6 +26,14 @@ server {
         proxy_pass $agreements_s3_url;
     }
 
+    # Hack to force g7 communications to the old bucket to preserve upload times
+    location ^~ /g-cloud-7-updates/communications {
```
TheDoubleK: Other rules in this file have a space between the `^` and `~` - I don't know enough about nginx config to say if that is significant at all.

robyoung: It's a different kind of rule. The ones that start with a `~` on it's own are regular expression rule. This one (`^~`) is a prefix rule that overrides regular expression rules.

http://nginx.org/en/docs/http/ngx_http_core_module.html#location

allait:
:+1:

PR #162, opened 2015-11-26
Update Google site verification code

robyoung:
Our access to Google Webmaster Tools is no longer working. After speaking with Ash we have decided to replacing the verification code.

We are reducing the TTL to 5 minutes in case there is any problem with replacing the old verification code.

See:
- https://github.com/alphagov/digitalmarketplace-aws/pull/163
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/189


allait:
:+1:

PR #163, opened 2015-11-26
Allow robots to crawl search results

robyoung:
The search results pages will be prevented from being indexed by a meta
tag on the pages themselves.

See:
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/189
- https://github.com/alphagov/digitalmarketplace-aws/pull/162


allait:
:+1:

PR #166, opened 2015-12-07
Mandrill DNS records

allait:
### Use TXT record instead of SPF record type
Mandrill SPF verification doesn't recognize current SPF value.

From Route 53 docs:

> ... we no longer recommend that you create resource record sets for
> which the record type is SPF ... its use is no longer appropriate for
> SPF version 1; implementations are not to use it ... Instead of an
> SPF record, we recommend that you create a TXT record

(http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html#SPFFormat)

### Add mandrill DKIM record and change SPF/DKIM TTL to 300
Setting a short TTL in case we need to revert the records. DKIM
value copied from Mandrill setup instructions page:
https://mandrillapp.com/settings/sending-domains?q=digitalmarketplace

### DIFF
--- a/cloudformation_templates/aws_route53_root_records.json
+++ b/cloudformation_templates/aws_route53_root_records.json
@@ -35,7 +35,22 @@
         "ResourceRecords": [
           "\"v=spf1 include:_spf.google.com include:spf.mandrillapp.com ~all\""
         ],
-        "TTL": "3600"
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Looks right to me :+1:

PR #167, opened 2015-12-07
Join SPF record with Google verification TXT record

allait:
Route53 can't create a second TXT root record resource since one
already was created by GoogleVerification resource. So instead we
just add the SPF record to the GoogleVerification resource.


TheDoubleK:
:+1:

PR #168, opened 2015-12-17
Prune old versions

robyoung:
## Add method to prune old Beanstalk app versions
There is a limit to the number of Beanstalk application versions that you can have, this limit is 500 per account and AWS refuses to increase it stating that we should be deleting old versions.

This adds a method to prune old Beanstalk application versions by listing them and deleting any above the number to keep.

## 	Prune old versions when releasing
When releasing to preview or staging (the only releases that create new versions) we need to prune the old versions before we create the version.

I have not added this to the `deploy` command because that is done manually and I don't think it should have unexpected side effects.

### DIFF
--- a/dmaws/deploy.py
+++ b/dmaws/deploy.py
@@ -65,6 +65,19 @@ class Deploy(object):
 
         return version_label, bool(created_version)
 
+    def prune_old_versions(self, number_to_keep):
+        assert isinstance(number_to_keep, (int, float))
+
+        versions = self.beanstalk.list_application_versions(self.eb_application)
+
+        count = 0
+        if len(versions) > number_to_keep:
+            for version in versions[number_to_keep:]:
+                count += 1
+                self.beanstalk.delete_application_version(self.eb_application, version['VersionLabel'])
+
+        return count
+
     def deploy(self, version_label):
         self.log("=== Deploying version %s to %s",
                  version_label, self.eb_environment)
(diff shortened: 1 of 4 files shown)


dmaws/deploy.py
```diff
@@ -65,6 +65,19 @@ def create_version(self, version_label, with_sha=False, description='',
 
         return version_label, bool(created_version)
 
+    def prune_old_versions(self, number_to_keep):
+        assert isinstance(number_to_keep, (int, float))
+
+        versions = self.beanstalk.list_application_versions(self.eb_application)
```
TheDoubleK: Is the order of versions that this returns guaranteed by `beanstalk`? It needs to be ordered with the oldest first for this method to work. (At least to avoid deleting recent versions.)

robyoung: It's returned with the most recent first and this will remove the tail of that list.

TheDoubleK: Ah yes, the "slice" notation - so terse it's easy to completely miss it. (At least on the bus early in the morning.)

TheDoubleK:
This looks good :+1: - difficult to be sure until we see it delete some versions in a real environment.  Are we going to set `DM_NUM_APPLICATIONS_TO_KEEP` in credentials, or just stick with the 20 default for now?

robyoung:
Yea, I was going to watch it run through on development to see it work.

PR #170, opened 2015-12-31
Add command to export database data

robyoung:
This adds a command to export data from the database of an environment cleaned up for a specific environment. The result of running this command is two `pg_dump` files (one for staging and one for preview) that can be loaded into another database.

## The basic process
- Create a new snapshot of database
- Restore a new instance from this snapshot
- Run SQL commands to scrub data for staging
- If environment is not 'staging' then run SQL commands to further scrub data for preview
- `pg_dump` and pipe directly into `psql` pointing at the target instance

## The data scrubbing
Most of the scrubbing is related to the requirement that no data related to an open framework should live anywhere other than production and we should avoid the risk of sending emails to users from anywhere other than production. I think that this demands more discussion as staging is considered equal to production and would benefit from having the same data in it.

### Staging
- Update all users to remove their name and email address
- Remove any draft services and supplier framework records for frameworks that are still open
- Remove any suppliers that are not associated with any framework. This leads on from the previous point, they are likely to be a supplier on an 'open' framework.
- Remove all audit events. These contain personal information and changes related so applications to 'open' frameworks peppered throughout. Unpicking it would be hard work and error prone.
- Blank out declarations leaving just the status as these contain sensitive information.

### Preview
- Update all passwords to 'Password1234' so that we can easily assume the role of accounts.

## Overwriting data
Importing this into another environment will overwrite the entire database resulting in anything manually entered or modified on that environment being lost. This is probably ok for staging but may be difficult on preview. I suggest that we don't automatically import on preview but rather do it on demand.


dmaws/commands/exportdata.py
```diff
@@ -0,0 +1,51 @@
...
+from ..rds import RDS, RDSPostgresClient
+
+
+@cli_command('exportdata', max_apps=0)
+@click.option('--export-path', default='.',
+              help="Path to write export files.")
+def exportdata_cmd(ctx, export_path):
+    plan = StackPlan.from_ctx(ctx, apps=['database_dev_access'])
+
+    if not plan.create(create_dependencies=False):
```
allait: Do we need dev access to production to take a snapshot or is this to avoid manually creating security groups for snapshot DB?

I'm a bit worried that the difference between cleaning up a snapshot and wiping out production data is `tmp_` in the `RDSPostgresClient` argument.

dmaws/commands/importdata.py
```diff
@@ -0,0 +1,31 @@
...
+@click.argument('export_file')
+def importdata_cmd(ctx, export_file):
+    """Import data into an environments database."""
+    plan = StackPlan.from_ctx(ctx, apps=['database_dev_access'])
+
+    if not plan.create(create_dependencies=False):
+        ctx.log("Failed to allow dev access", color='red')
+        sys.exit(1)
+
+    plan.info(['database'])
```
allait: I think we might be able to skip this, since creating dev access plan should gather database info and fill variable values anyway.

dmaws/rds.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def get_instance(self, url=None, id=None):
+        for instance in self.conn.get_all_dbinstances():
+            if url and instance.endpoint[0] == url.split(':')[0]:
+                return instance
+            if id and instance.id == id:
+                return instance
+
+    def create_new_snapshot(self, snapshot_id, instance_id):
+        """Create a new RDS snapshot deleting the existing one if there
```
allait: I think you accidentally a word

dmaws/rds.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        try:
+            self.conn.get_all_dbsnapshots(snapshot_id)
```
allait: I suspect it's a boto thing, but just to check: why do we need to get all snapshots if we try to delete the conflicting one either way?

dmaws/rds.py
```diff
@@ -0,0 +1,206 @@
...
+            self.conn.delete_dbsnapshot(snapshot_id)
```
allait: Could be `self.delete_snapshot`?

dmaws/rds.py
```diff
@@ -0,0 +1,206 @@
...
+        ], logger=self.log, ignore_errors=True)
+
+    def get_open_framework_ids(self):
+        self.cursor.execute("SELECT id FROM frameworks WHERE status = 'open'")
+        return tuple(framework[0] for framework in self.cursor.fetchall())
+
+    def clean_database_for_staging(self):
+        self.log("Update users")
+        self.cursor.execute(
+            "UPDATE users SET name = %s, email_address = id || '@example.com'",
```
allait: This works pretty well for staging, but for `preview` using user id makes it harder to find out what user account can be used for each supplier. For preview, can we do something similar to what API to API script is doing now: instead of importing existing users create a `suppier-<supplier_id>@user.dmdev` user for each active supplier (and change the supplier contact details to match that email).

dmaws/rds.py
```diff
@@ -0,0 +1,206 @@
...
+        self.log("Delete draft services")
+        self.cursor.execute(
+            "DELETE FROM draft_services WHERE framework_id IN %s",
+            (open_framework_ids,))
+        self.log("Delete supplier frameworks")
+        self.cursor.execute(
+            "DELETE FROM supplier_frameworks WHERE framework_id IN %s",
+            (open_framework_ids,))
+
+        # Fix suppliers with services but no supplier_framework
```
allait: I'm a bit lost about what this does. We're deleting supplier_frameworks for open frameworks (which shouldn't have live services). So that doesn't affect the live frameworks. This query then tries to create records for existing live services, which should already exist in production.

This also checks for supplier ids without any records, but what we're actually interested is whether the `(supplier_id, framework_id)` pair already exists.

allait:
Trying to summarise our offline conversation:

We need to make sure that no sensitive artefacts (ie staging database dump files) are left behind after the job runs, so we might join export and import into a single command that cleans up any created files. This could also allow us to stream from export DB to the target DB instead of using a temporary dump file.

The command should also allow running staging and preview transfer jobs separately.

dmaws/rds.py
robyoung: No, the sentence is as intended. If it's not clear I can make it more explicit.
> Create a new RDS snapshot deleting the existing snapshot if there is one.

dmaws/rds.py
robyoung: Yes, this is how you get a specific db snapshot. If it doesn't exist it will raise a 404 short circuiting the delete.

dmaws/rds.py
robyoung: Yes, good call.

dmaws/rds.py
robyoung: On preview we don't have `supplier_frameworks` records for G-Cloud 5 (I haven't checked if this is the case on production). We remove any suppliers that don't have any `supplier_frameworks` record on the assumption that they're only associated with open frameworks. This statement adds in records for suppliers that have services to avoid us trying to delete those suppliers.

The deletion of dangling suppliers has changed since this was done though so it might warrant another look. Maybe building the list of dangling suppliers my checking both `supplier_frameworks` and `services`.

allait: The data on preview will get replaced though, what matters is that `supplier_frameworks` records for all live frameworks are not affected by the deletion of open frameworks. If that's the case then whatever live supplier frameworks exist (or not) in production should be fine for staging and preview (assuming production data is correct, otherwise we need to fix that first).

dmaws/commands/exportdata.py
robyoung: To the first question; yes. Although that's a good point, it would be good to handle that in the RDS client, I'll take a look at it.

dmaws/context.py
```diff
@@ -19,6 +20,18 @@ def __init__(self, home=os.getcwd(), dry_run=False):
...
         self.create_dependencies = False
 
+    def new_context(self, stage, environment, vars_files):
+        """Create a new context with different variables"""
+        ctx = Context()
+        ctx.stage = stage
+        ctx.environment = environment
+        ctx.load_variables(
+            files=ctx.get_variables_files(True, vars_files))
+        ctx.stacks = copy.deepcopy(self.stacks)
```
allait: With stacks replaced by 'built stacks' on the original context I'm not really sure whether we can be sure that this new context doesn't have any data carried over from the original environment. I think it might be better to do a fresh `ctx.load_stacks(stacks_file)`, unless I'm missing a reason why we need a copy.

dmaws/commands/migratedata.py
```diff
@@ -0,0 +1,90 @@
...
+
+    target_ctx = ctx.new_context(
+        stage=target_stage,
+        environment=target_environment,
+        vars_files=[target_vars_file])
+
+    rds, pg_client = create_scrubbed_instance(ctx, target_stage)
+    dump_to_target(target_ctx, pg_client)
+
+    pg_client.close()
```
allait: `src_pg_client` seems to be closed in `dump_to_target`

dmaws/commands/migratedata.py
```diff
@@ -0,0 +1,90 @@
...
+    plan.info()
+
+    instance = rds.get_instance(plan.get_value('stacks.database.outputs')['URL'])
+
+    instance = rds.allow_access_to_instance(
+        instance, IMPORT_SECURITY_GROUP_NAME,
+        ctx.variables['dev_user_ips'],
+        ctx.variables['vpc_id'])
+
+    pg_client = RDSPostgresClient.from_boto(
```
allait: Do we need to explicitly close this connection the same way we close `src_pg_client`?

dmaws/context.py
robyoung: The reason we don't do a fresh `ctx.load_stacks(stacks_file)` is that we don't have the `stacks_file` at the point that this is used (in the `migratedata` command). The reason we need a new context is to create and delete stacks in two different environments at the same time. 

If you think might be a risk it may be worth while just biting the bullet and getting all the construction code out of `dmaws.cli.py`. What do you think?

allait: We can store the `stacks_file` in `ctx.stacks_file`, since we'd still need to find a way to pass it in even if we move the logic from `@cli_command`. And most commands still need one context, so I'm not sure if moving it out is worth it right now.

dmaws/commands/migratedata.py
robyoung: Nice spot, that's a bug.

allait:
:+1:

PR #171, opened 2015-12-31
Do not index the assets subdomain

robyoung:
[#109802160](https://www.pivotaltracker.com/story/show/109802160)


allait:
Looks good! :+1: 
Should we do the new AMI right here?

robyoung:
err, no? ... sorry, I got :+1: blindness and didn't spot the AMI comment. I'll add that in a separate PR.

PR #172, opened 2016-01-04
Add bucket env variables to admin frontend

robyoung:
The communications bucket is required uploading communications. The
others have been added now as they will be useful soon and it's a bit
confusing to debug when it's not available.

### DIFF
--- a/stacks.yml
+++ b/stacks.yml
@@ -260,6 +260,8 @@ admin_frontend:
   template: cloudformation_templates/aws_digitalmarketplace_admin_frontend.json
   dependencies:
     - documents_s3
+    - communications_s3
+    - submissions_s3
     - agreements_s3
     - api
     - admin_frontend_app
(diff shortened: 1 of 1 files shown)


allait:
I think buckets need to be added to the stack dependencies, similar to how we do it in the supplier app stack

stacks.yml
```diff
@@ -277,6 +277,9 @@ admin_frontend:
     EnvVarDmApiUrl: "{{ stacks.api.outputs.URL }}"
     EnvVarDmS3DocumentBucket: "{{ stacks.documents_s3.outputs.Name }}"
```
TheDoubleK: Is the `EnvVarDmS3DocumentBucket` obsolete now? If so it would be good to remove it to avoid clutter.

robyoung: It is at the moment but once this change goes in the admin app can be updated and then `EnvVarDmS3DocumentBucket` can be removed.

playbooks/roles/cloudwatch/defaults/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-aws_log_agent_url: https://s3.amazonaws.com/aws-cloudwatch/downloads/latest/awslogs-agent-setup.py
+aws_log_agent_url: http://s3.amazonaws.com/aws-cloudwatch/downloads/latest/awslogs-agent-setup.py
```
allait: I suppose it's about time we gave up on this getting fixed

robyoung:
:tada:

allait:
:+1:

PR #174, opened 2016-01-05
Add SPF DNS record to allow sending emails from Zendesk

allait:
Adding to existing Google and Mandrill SPF records.
(https://support.zendesk.com/hc/en-us/articles/203683886-SPF-Learn-More)


robyoung:
:+1:

PR #175, opened 2016-01-05
Update environment variable name to the one used in the application

TheDoubleK:
This updates the environment variable to be the one used in the supplier app here:
https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/config.py#L49
and here:
https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/frameworks.py#L400

Depends on new variable name in:
 - [x] https://github.gds/gds/digitalmarketplace-credentials/pull/54/files


allait:
:+1:

PR #176, opened 2016-01-08
Add DB parameter group to enable slow query logging

allait:
Sets a Parameter Group on DB instances than can be used
to tweak database settings. Dynamic parameters can be
then changed without interruption.

Enables logging of queries >1000ms on preview by default.
Logging is disabled for staging and production.

**Note:** even though AWS docs state that 
> The parameter group name itself is changed immediately, but the actual parameter changes are not applied until you reboot the instance without failover. The db instance will NOT be rebooted automatically and the parameter changes will NOT be applied during the next maintenance window.
(http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBInstance.html)

CloudFormation seems to have restarted the instance when this was applied to preview causing a <1 min downtime. If we decide to merge this and the same happens in staging we'll have to either revert or apply this to production when we can deal with a short service interruption 🔥

### DIFF
--- a/cloudformation_templates/aws_rds_database.json
+++ b/cloudformation_templates/aws_rds_database.json
@@ -36,6 +36,10 @@
     "DBSnapshotIdentifier": {
       "Type": "String",
       "Description": "The identifier for the DB snapshot to restore from."
+    },
+    "SlowQueryLogDurationMs": {
+      "Type": "Number",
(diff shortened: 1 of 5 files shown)


TheDoubleK:
Looks good :+1:

PR #177, opened 2016-01-27
Delete all draft services when migrating

robyoung:
We do not ever need draft services to be copied over to staging. If they
are for an open or pending framework we definitely don't want them for
legal reasons. If they are for a live framework  they are not very
useful.

Supplier framework records need to be removed for open or pending
frameworks for legal reasons.


allait:
:+1:

PR #180, opened 2016-01-27
Update nginx AMI with nginx 1.8.1

allait:
There's a new stable nginx release with some security fixes related
to the resolver (which we're using to connect to application ELBs)

http://nginx.org/en/CHANGES-1.8


TheDoubleK:
:+1: 
(I found the new AMI without needing to ask where the number comes from - I'm learning at last!)

PR #181, opened 2016-01-28
Add service role for ElasticBeanstalk health reporting

allait:
> The following statement contains all of the permissions that Elastic
> Beanstalk needs to monitor environment health

http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts-roles.html#concepts-roles-service
http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/health-enhanced-api.html

### DIFF
--- a/cloudformation_templates/app_base.j2
+++ b/cloudformation_templates/app_base.j2
@@ -124,6 +124,54 @@
       }
     },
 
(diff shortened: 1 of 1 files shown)


robyoung:
:+1:

PR #182, opened 2016-02-03
Add login and Mandrill secrets to buyer app

TheDoubleK:
The buyer app will need access to these tokens for processing logins and sending emails.

So this change will pull in the new `shared_tokens` from this related PR:
https://github.gds/gds/digitalmarketplace-credentials/pull/56

I've also removed the `admin_password_hash` which is no longer used in the admin app at all.


robyoung:
:+1:

PR #183, opened 2016-02-05
S3 versioning and EB IAM delete permission

allait:
### Enable Object Versioning for S3 buckets
Object versioning will automatically keep old file versions so
that there's less need for us to do manual copies and backups
of modified or removed documents.

### Add s3:Delete* permission for EB InstanceProfile
Admin app sends delete requests to remove countersigned agreements,
so we need to add a permission to EB instance profile policy.


robyoung:
:+1:

PR #184, opened 2016-02-05
Provide SECRET_KEY env var for apps

robyoung:
Currently we use different environment variable names for the Flask secret key and then override it with `os.environ` in the [root Config class](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/config.py#L52).

This is problematic for a few reasons:

1. `SECRET_KEY` is a Flask thing and naming it something different (even
   if it is more descriptive) obfuscates what it's being used for.
2. Setting it from the environment in the base class means that we
   cannot set a default for it in one of the environment specific child
   classes.

If we think that it is more helpful to have more descriptive names rather than stay consistent with Flask's naming than we should map the name we use to `SECRET_KEY` in `init_app`.


TheDoubleK:
:+1:

PR #185, opened 2016-03-22
Enable gzip for css and javascript

robyoung:
By default nginx only compresses text/html this change also compresses
css and javascript.


allait:
:+1:

PR #187, opened 2016-04-11
Update preview TSL certificate ARN

allait:
Updates certificate name to match the new letsencryp certificate.
Letsencrypt update job will always rename the new certificate to
keep the current name, so future renewals should require no more
changes.

Removes iam_certificate_id since it appears to be unused now.


TheDoubleK:
:+1:

PR #188, opened 2016-04-12
Set generic password for all users in migration to staging

TheDoubleK:
It will be useful for testing things to be able to log in as different users on staging.


allait:
👍

PR #189, opened 2016-05-11
Update staging and production TLS certificate

allait:
New certificate is uploaded to IAM so we can update the nginx ELBs to use it instead of the expiring one.


TheDoubleK:
:+1:

PR #190, opened 2016-05-18
Update RDS database to PostgreSQL 9.4

allait:
Changes the database engine version to 9.4. Major upgrades can only be done one at a time, so there's no way to move from 9.3 to 9.5 in one step. "allowMajorUpgrades" has to be set to true to apply the change.

Despite the description in the [AWS docs][1], there doesn't seem to be a way to change the DBParameterGroup Family parameter. AWS sets the resource status to 'UPDATE_COMPLETE', but the parameter group still has the old database version. And there doesn't seem to be a way to update the parameter group family using AWS CLI, so it's likely an error in the CloudFormation docs.

The updated database version however requires a matching parameter group family, so we rename the parameter group resource to force CloudFormation to replace the old group with a new one. This runs the expected update, removing the old group once the DB instance has finished updating, but requires a manual change to the template each time.

[1]: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-dbparametergroup.html#cfn-rds-dbparametergroup-family

### DIFF
--- a/cloudformation_templates/aws_rds_database.json
+++ b/cloudformation_templates/aws_rds_database.json
@@ -7,6 +7,10 @@
       "Type": "String",
       "Description": "Database name"
     },
+    "DBVersion": {
+      "Type": "String",
+      "Description": "Database version number"
+    },
     "DBUser": {
       "NoEcho": "true",
       "Type": "String",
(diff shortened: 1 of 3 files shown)


allait:
I've tested this on a separate database-only environment. I'm going to merge this and try to get it to preview/staging today.

TheDoubleK:
:+1:

PR #191, opened 2016-06-14
Increase instance size for API and supplier app

TheDoubleK:
Since we expect an increase in requests for the last days of the G-Cloud 8 service submission, we raise the size of EC2 instances for supplier frontend and Data API apps.

This is the same as we did for G7 and it was all good (see Alexey's commit here: https://github.com/alphagov/digitalmarketplace-aws/commit/df1f1effc792f5d7211896e1dcd8ea8ae1609dcc)
We can reduce these to the previous value after the service submission is closed.

### DIFF
--- a/cloudformation_templates/aws_rds_database.json
+++ b/cloudformation_templates/aws_rds_database.json
@@ -7,10 +7,6 @@
       "Type": "String",
       "Description": "Database name"
     },
-    "DBVersion": {
-      "Type": "String",
-      "Description": "Database version number"
-    },
     "DBUser": {
       "NoEcho": "true",
       "Type": "String",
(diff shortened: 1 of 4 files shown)


pcraig3:
Yeah, looks good. 👍 

Are we going to lower the `max_instance_count` to `5`, or is it on `10` forever now?

TheDoubleK:
It didn't ever get rolled back to 5, not sure if that was on purpose or not.  But if we do want to go back to 5 then now isn't the time.

pcraig3:
Latest commit [looks like it's doing what it's meant to](https://github.com/alphagov/digitalmarketplace-aws/compare/0e3db3e9dd5431bcd233d5dff027c827b0d0fd0d...6efd5d1d2748a44446e692969f47981da085395c). 👍  👍

TheDoubleK:
Oooh, nice way to compare.

Bit nervous, but going to run this and see what happens.

PR #193, opened 2016-06-24
Revert "Increase instance size for API and supplier app"

allait:
Service submissions are closed and the request spike is over, so we can scale down the instance sizes. We've reached the 10 instance autoscaling limit yesterday, but it's back to 3 instances for API and supplier frontend now.


TheDoubleK:
:+1:

TheDoubleK:
We'll do this now before the Postgres upgrade goes in I guess?

PR #194, opened 2016-06-30
Kibana log streaming

allait:
### Remove dependency on monitoring from database stack
Database stack doesn't reference the monitoring output, so there's no reason to have an explicit dependency. The only connection between RDS and monitoring is the notification when a DB instance is deleted, but there's little value in them for standalone databases.

Monitoring is still required for apps, so there's no risk of accidentally skipping the stack update.

### Abort create process if one of the dependent stack fails to create

Stops the creation process if a stack fails to create. This is similar to how we stopped when a dependant stack didn't exist, but for some reason there was no similar check for creating dependencies.

### Fix a race condition when creating monitoring metrics

Using JSONLogGroup reference to get the log group name instead of creating the name string forces cloudformation to wait until the log group resource has finished creating. Without the dependency both log group and metric creation are started at the same time and metric fails since the log group name doesn't exist yet.

### Move param_to_env to utils
param_to_env captures our naming convention for parameter name transformation, so it could be used by other methods/commands.

## Cloudwatch log streaming

screen shot 2016-06-30 at 17 12 04

### Add kibana stack to create AWS Elasticsearch service for logs
Creates an AWS Elasticsearch instance with IP access restriction for developers.

### Add AWS Lambda CloudFormation template and stacks
There are two ways to create a Lambda function using CloudFormation:
* by inlining the code into the template directly, which is limited to 4096 bytes
* from an S3 object containing a ZIP archive with the function code

Since the inline size is quite restrictive we're going with the S3 route. The stacks will create an S3 bucket and then attempt to create a Lambda function using the `S3Key` object that should be stored in that bucket. Since there's no way to upload a file to S3 using CloudFormation that file has to be uploaded manually (after the bucket stack was created, but before the Lambda one) when the stack is run for the first time.

Rerunning the stack if the S3 file was changed won't update the Lambda code, since CloudFormation considers the template to be unchanged. There's a way to add S3 object version to the resource (or use manual versioning by changing the file name for each version) forcing CloudFormation to apply the update, but that requires changing the variable for each new function version (similar to how packer AMI
flow works right now). Instead, since code archive creation will require additional steps anyway, we're going to add a command to release and apply a new lambda version.

### Add lambda-release dmaws command to push lambda code updates
Since there's no convenient way to upload Lambda function code with CloudFormation we need a separate command to create the code archive, upload it to S3 and update the Lambda.

`lambda-release` command takes the name of the lambda function stack and:
1. Looks for the function code in `lambdas/<lambda_name>`. `lambda_name` is the name of the stack with `"_lambda"` part removed. The folder should contain a main.js file (since this is what our runtime and handler parameters in aws_lambda template are set to).
2. Create the environment file with all `EnvVar` parameters stored in JSON.
3. Archive the code and upload it to `S3Bucket/S3Key`
4. Update the Lambda function code with the new file from S3

Environment variables are taken from the stack parameters matching the `EnvVar` naming convention. Parameter names are rewritten using `param_to_env`, similar to how we do it with ElasticBeanstalk app environment variables. The file is stored as `'env'` and added to the code package, so that it can be loaded by the function code.

Since all CloudFormation parameters need to be declared by the template, even though they're not used by any of the resources, we have to generate the EnvVar declarations using Jinja, same as in the app_base template.

We could update function code directly instead of uploading to S3 first, but this approach makes it possible to recreate CloudFormation stack without losing the function code. Since CloudFormation references the same S3 object it will recreate the stack with the latest function code.
The downside of this approach is that we need to upload a file to S3 before the first lambda stack run. The alternative would be to use an empty function code block in the CloudFormation template and then upload the actual code using the new command.

### Add cloudwatch_logs Lambda function source code
Based on the function created when setting up streaming to Amazon Elasticsearch service.

Code is modified to read Elasticsearch endpoint from the environment file and adds some field changes specific to our log format in `buildSource` (rewriting request times so that apache and nginx formats match and dropping time fields that confused Kibana).

### Add log streaming from CloudWatch to Lambda function
Creates a log subscription for json log group. There can be only one subscription per log group, which we set up with an empty filter, so that all log streams are sent to the lambda function, which will then store them in Elasticsearch. This is not ideal, since our log group contains both apache and application logs
and with one subscription the only way to split them is inside the lambda function itself. And there's no way to set up separate functions for more granular filters (eg notifying on exceptions).

### Point *.logs.beta domains to Jenkins instance IP
Allows proxying Elasticsearch instances through Jenkins nginx with more useful domain names and added authentication.

### Don't stream ELB health check request logs to Kibana
Filters out /_status?ignore-dependencies logs from streaming to the AWS Lambda function. Health check logs aren't really useful since they're mostly the same and removing them makes it easier to see actual user requests.

### Make Kibana Elasticsearch cluster only accessible from Jenkins
Since there's no way to set up user authentication on AWS ES service we make the Elasticsearch server accessible only from the Jenkins IP and set up a proxy with authorization on the Jenkins instance. This way ES is protected both by an IP restriction applied to the Jenkins nginx port and the authorization set up in nginx.

### Add a stack to open Jenkins port for letsencrypt checks
When verifying the domain ownership letsencrypt CA needs to be able to make a request to the domain server. Since the source IP address can change often we open the HTTP port for all IPs and change nginx config on Jenkins instance to only listen on 443, which means our proxy is still restricted to our developer IPs and port 80 is free to be used by letsencrypt standalone server when required.

The stack can be safely removed when unsused since it's only needed for the verification when requesting the certificate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,23 @@ Packer and cloudformation won't automatically remove old AMI versions, so this h
 done manually. AMIs can be deleted from the EC2 console by deregestering the AMI and
 deleting the related EBS snapshot.
 
+## Deploying AWS Lambda functions
+
+Lambda functions are created by the CloudFormation stack, but rerunning the stack won't updated
+the function code. Instead, functions can be updating using the `lambda-release` command:
+
+```
+dmaws preview preview lambda-release cloudwatch_logs_lambda
+```
+
+This will package function code from `lambdas/cloudwatch_logs`, upload the archive to S3 and update
+the function code in AWS Lambda.
+
+This command can also be used to upload the initial function code archive to S3 before running the
+function stack for the first time. In this case, updating the function code will fail, since the
+function doesn't exist yet, but an archive will be uploaded to S3 and CloudFormation will use it
+when creating the function stack.
+
 ## SSHing into instances
 
 You can SSH onto instaces using the private key set up previously (SSH key pair).
--- a/cloudformation_templates/aws_jenkins.json
+++ b/cloudformation_templates/aws_jenkins.json
@@ -17,7 +17,11 @@
     },
     "Subdomain": {
       "Type": "String",
-      "Description": "Subdomain to use"
+      "Description": "Jenkins subdomain"
+    },
+    "LogsSubdomain": {
+      "Type": "String",
+      "Description": "Kibana proxy subdomain"
     },
     "HostedZoneName": {
       "Type": "String",
@@ -122,13 +126,24 @@
       }
     },
 
-    "Route53RecordSet": {
+    "CIRoute53RecordSet": {
       "Type": "AWS::Route53::RecordSet",
       "Properties": {
         "HostedZoneName": {"Ref": "HostedZoneName"},
-        "Name": {"Fn::Join": ["",[
-          {"Ref": "Subdomain"}, ".", {"Ref": "RootDomain"}, "."
-         ]]},
+        "Name": {"Fn::Join": [".", [{"Ref": "Subdomain"}, {"Ref": "HostedZoneName"}]]},
+        "Type": "A",
+        "ResourceRecords": [
+          {"Ref": "ElasticIP"}
+        ],
+        "TTL": "300"
+      }
+    },
+
+    "LogsRoute53RecordSet": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
+        "HostedZoneName": {"Ref": "HostedZoneName"},
+        "Name": {"Fn::Join": [".", ["*", {"Ref": "LogsSubdomain"}, {"Ref": "HostedZoneName"}]]},
         "Type": "A",
         "ResourceRecords": [
           {"Ref": "ElasticIP"}
--- a/cloudformation_templates/aws_kibana.json
+++ b/cloudformation_templates/aws_kibana.json
@@ -1,0 +1,63 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Description": "Kibana AWS Elasticsearch domain with CloudWatch streaming",
+  "Parameters": {
+    "DomainName": {
+      "Type": "String",
+      "Description": "Domain name for AWS Elasticsearch service"
+    },
+    "UserIPs": {
+      "Type": "CommaDelimitedList",
+      "Description": "List of IP addresses that can access the Elasticsearch domain"
+    }
+  },
+
+  "Resources": {
+    "ElasticsearchDomain": {
+      "Type": "AWS::Elasticsearch::Domain",
+      "Properties": {
+        "DomainName": {"Ref": "DomainName"},
+        "EBSOptions": {
+          "EBSEnabled": true,
+          "VolumeSize": 15,
+          "VolumeType": "gp2"
+        },
+        "ElasticsearchClusterConfig": {
+          "InstanceType": "t2.micro.elasticsearch",
+          "InstanceCount": 1
+        },
+        "AccessPolicies": {
+          "Version": "2012-10-17",
+          "Statement": [{
+            "Sid": "",
+            "Effect": "Allow",
+            "Principal": {"AWS": "*"},
+            "Action": "es:*",
+            "Resource": "arn:aws:es:*",
+            "Condition": {
+              "IpAddress": {
+                "aws:SourceIp": {"Ref": "UserIPs"}
+              }
+            }
+          }]
+        }
+      }
+    }
+  },
+  "Outputs": {
+    "Name": {
+      "Description": "Elasticsearch domain name",
+      "Value": {"Ref": "ElasticsearchDomain"}
+    },
+    "Domain": {
+      "Description": "Elasticsearch domain endpoint",
+      "Value": {"Fn::GetAtt": ["ElasticsearchDomain", "DomainEndpoint"]}
+    },
+    "URL" : {
+      "Value" : {"Fn::Join": ["", [
+        "https://", {"Fn::GetAtt": ["ElasticsearchDomain", "DomainEndpoint"]}
+      ]]},
+      "Description" : "Elasticsearch domain URL"
+    }
+  }
+}
--- a/lambdas/cloudwatch_logs/main.js
+++ b/lambdas/cloudwatch_logs/main.js
@@ -1,0 +1,239 @@
+// Based on AWS Cloudwatch "Stream to Amazon Elasticsearch Service" funcion v1.1.2
+// Required stack parameters:
+// * EnvVarElasticsearchEndpoint - Amazon Elasticsearch Service domain endpoint
+
+var https = require('https');
+var zlib = require('zlib');
+var crypto = require('crypto');
+
+var environment = JSON.parse(require('fs').readFileSync('env', 'utf8'));
+var endpoint = environment['ELASTICSEARCH_ENDPOINT'];
+
+exports.handler = function(input, context) {
+    // decode input from base64
+    var zippedInput = new Buffer(input.awslogs.data, 'base64');
+
+    // decompress the input
+    zlib.gunzip(zippedInput, function(error, buffer) {
+        if (error) { context.fail(error); return; }
+
+        // parse the input from JSON
+        var awslogsData = JSON.parse(buffer.toString('utf8'));
+
+        // transform the input to Elasticsearch documents
+        var elasticsearchBulkData = transform(awslogsData);
+
+        // skip control messages
+        if (!elasticsearchBulkData) {
+            console.log('Received a control message');
+            context.succeed('Control message handled successfully');
+            return;
+        }
+
+        // post documents to the Amazon Elasticsearch Service
+        post(elasticsearchBulkData, function(error, success, statusCode, failedItems) {
+            console.log('Response: ' + JSON.stringify({
+                "statusCode": statusCode
+            }));
+
+            if (error) {
+                console.log('Error: ' + JSON.stringify(error, null, 2));
+
+                if (failedItems && failedItems.length > 0) {
+                    console.log("Failed Items: " +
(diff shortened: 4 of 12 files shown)


cloudformation_templates/aws_kibana.json
```diff
@@ -0,0 +1,63 @@
...
+
+  "Resources": {
+    "ElasticsearchDomain": {
+      "Type": "AWS::Elasticsearch::Domain",
+      "Properties": {
+        "DomainName": {"Ref": "DomainName"},
+        "EBSOptions": {
+          "EBSEnabled": true,
+          "VolumeSize": 15,
+          "VolumeType": "gp2"
```
TheDoubleK: Might `st1` instances be OK for this?  They're half the price of `gp2`, and use cases include "log processing" (https://aws.amazon.com/ebs/details/).

cloudformation_templates/aws_kibana.json
```diff
@@ -0,0 +1,63 @@
...
+        },
+        "ElasticsearchClusterConfig": {
+          "InstanceType": "t2.micro.elasticsearch",
+          "InstanceCount": 1
```
TheDoubleK: Isn't three instances the minimum recommended for an elasticsearch cluster?

cloudformation_templates/aws_kibana.json
allait: It's a $1/month difference for SSD instead of HDD. And I think `st1` has a 500Gb volume size minimum.

TheDoubleK:
The S3 route for storing and updating lambda functions seems pretty roundabout.
Is the code likely to change very often? Is it likely to exceed 4096 bytes (or does it already)? If no then it might be worth going with the inline version to start with?

cloudformation_templates/aws_kibana.json
allait: For a production cluster I'd say yes – it's the minimum HA set up. But in this case I don't think we care that much if we lose the data or the instance goes down for a short time, so I went with the cheapest possible option to begin with.

allait:
@TheDoubleK it does exceed 4096 bytes (the default function AWS creates for streaming is about 7k)

lambdas/cloudwatch_logs/main.js
```diff
@@ -0,0 +1,239 @@
...
+
+        // parse the input from JSON
+        var awslogsData = JSON.parse(buffer.toString('utf8'));
+
+        // transform the input to Elasticsearch documents
+        var elasticsearchBulkData = transform(awslogsData);
+
+        // skip control messages
+        if (!elasticsearchBulkData) {
+            console.log('Received a control message');
```
TheDoubleK: Where do these `console.log` messages end up? Can we ever get to see them?

allait: It creates a new log group in CloudWatch for each function. It's a bit annoying since I don't think there's a way to configure it and it sets them to never expire by default. But I don't think the functions write very much, and I set a shorter expiration for preview manually.

allait: It also has monitoring for the function runs and failures, so you can see the stats in Lambda interface and go to CloudWatch from there.

lambdas/cloudwatch_logs/main.js
```diff
@@ -0,0 +1,239 @@
...
+
+    var bulkRequestBody = '';
+
+    payload.logEvents.forEach(function(logEvent) {
+        var timestamp = new Date(1 * logEvent.timestamp);
+
+        // index name format: cwl-YYYY.MM.DD
+        var indexName = [
+            'cwl-' + timestamp.getUTCFullYear(),              // year
+            ('0' + (timestamp.getUTCMonth() + 1)).slice(-2),  // month
```
TheDoubleK: Might be worth having a comment in the code about why we need `.slice(-2)` here as it seems a pretty funky thing to do. (I guess it is because of this? http://stackoverflow.com/questions/3605214/javascript-add-leading-zeroes-to-date)

lambdas/cloudwatch_logs/main.js
```diff
@@ -0,0 +1,239 @@
...
+
+function isValidJson(message) {
+    try {
+        JSON.parse(message);
+    } catch (e) { return false; }
+    return true;
+}
+
+function isNumeric(n) {
+    return !isNaN(parseFloat(n)) && isFinite(n);
```
TheDoubleK: The `isFinite` check seems redundant if it can be parsed as a float?

lambdas/cloudwatch_logs/main.js
allait: I haven't really read most of this to be honest 😄 This is the function that AWS console creates when you set up streaming to ES through the web interface. I extended `buildSource` a bit to clean up some repeated fields and drop support for non-JSON logs, but apart from that this is as-is, which might make it easier to update if AWS makes a newer version. (Although I'm not sure if we'd want to and I don't know how we'd find out).

lambdas/cloudwatch_logs/main.js
allait:
```javascript
> parseFloat(1/0)
Infinity
```
¯\\\_(ツ)_/¯

README.md
```diff
@@ -63,6 +63,23 @@ Packer and cloudformation won't automatically remove old AMI versions, so this h
 done manually. AMIs can be deleted from the EC2 console by deregestering the AMI and
 deleting the related EBS snapshot.
 
+## Deploying AWS Lambda functions
+
+Lambda functions are created by the CloudFormation stack, but rerunning the stack won't updated
+the function code. Instead, functions can be updating using the `lambda-release` command:
```
TheDoubleK: Is this something that will generally be done via a Jenkins job rather than from local command line?  If so we could say that in the README too, to make it clear.

cloudformation_templates/aws_jenkins.json
```diff
@@ -122,13 +126,24 @@
...
+    },
+
+    "LogsRoute53RecordSet": {
       "Type": "AWS::Route53::RecordSet",
       "Properties": {
         "HostedZoneName": {"Ref": "HostedZoneName"},
-        "Name": {"Fn::Join": ["",[
-          {"Ref": "Subdomain"}, ".", {"Ref": "RootDomain"}, "."
-         ]]},
+        "Name": {"Fn::Join": [".", ["*", {"Ref": "LogsSubdomain"}, {"Ref": "HostedZoneName"}]]},
```
TheDoubleK: The commit message says "*.logs.beta domains" - is `stacks.route53zone.outputs.HostedZoneName` always going to evaluate to "beta"?

stacks.yml
```diff
@@ -529,7 +529,7 @@ lambda_logs_subscription:
     - cloudwatch_logs_lambda
   parameters:
     Lambda: "{{ stacks.cloudwatch_logs_lambda.outputs.Arn }}"
-    FilterPattern: ''
+    FilterPattern: '{ $.request != "*/_status?ignore-dependencies *" || $.url != "*/_status?ignore-dependencies" }'
```
TheDoubleK: We might want to expand this to all `/_status` requests?  These will all generally be the "Apps are up" smoketests.  Not sure if there's anything else we can obviously throw away at this stage... probably keep it all for now.

stacks.yml
```diff
@@ -543,6 +543,17 @@ jenkins_ssh_access:
     ToPort: 22
     SecurityGroup: "{{ stacks.jenkins.outputs.InstanceSecurityGroup }}"
 
+jenkins_letsencrypt_access:
```
TheDoubleK: I'm not sure I follow what this is for - is this to get another new cert?  And then we kill it once they have verified the server and issued the cert to us?  Would we then need to turn it back on for renewal in a couple of years time?   Or have I misunderstood?

cloudformation_templates/aws_kibana.json
TheDoubleK: OK cool.  500Gb is probably overkill!

lambdas/cloudwatch_logs/main.js
TheDoubleK: LOL, remind me never to work with Javascript.

cloudformation_templates/aws_kibana.json
TheDoubleK: Seems reasonable.  Will we also have everything in Cloudwatch still, or are we going to kill Cloudwatch in favour of Kibana?

TheDoubleK:
Sorry for all the questions, just to help me get my head round it all a bit really.
I can't pretend I full grasp all the nuance of some bits, but in that it works on preview and nothing here is obviously wrong to me... :+1:

cloudformation_templates/aws_kibana.json
allait: We do need CloudWatch – it's where instances send logs to begin with, so streaming is for CloudWatch to Kibana. Also it's a more robust permanent storage than an ES cluster, so the idea is to keep only a few weeks of recent data in Kibana to simplify debugging/monitoring.

README.md
allait: Could be either. For now, I've done this locally, but we could set up a Jenkins job to deploy new changes after review.

cloudformation_templates/aws_jenkins.json
allait: Right now it is, and it's set in the stack variables, so we'd need another commit if we decide to change it to something else.

stacks.yml
allait: Letsencrypt creates a standalone server for a few minutes when it updates a certificate, so we do need it open every few months.

I went back and forth on keeping it open all the time. What I've done is move all nginx stuff to 443, so Jenkins and log proxies are only available via HTTPS and the security group for 443 only allows  AH IPs. Which should make it safe to open port 80 for everyone, so that we can use it for renewing letsencrypt certs. We could close it when we don't need it to avoid accidental nginx misconfiguration, but apart from that it shouldn't make any difference since nothing is listening most of the time.

PR #195, opened 2016-07-04
DSS domain redirect

allait:
### Add a stack to create DSS Route53 records
Points [www.]digitalservicesstore.service.gov.uk to the nginx ELB
using Route53 alias records.

There's a stage check for hosted zone domain value that will make
stack creation fail on non-production environments. This is making
sure that records don't get accidentally changed to staging ELB.

### DIFF
--- a/cloudformation_templates/aws_route53_digital_services_store.json
+++ b/cloudformation_templates/aws_route53_digital_services_store.json
@@ -1,0 +1,61 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Description": "A public hosted zone for the given root domain.",
+
+  "Parameters": {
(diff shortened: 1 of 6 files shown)


cloudformation_templates/aws_route53_digital_services_store.json
```diff
@@ -0,0 +1,61 @@
+{
```
TheDoubleK: This file doesn't contain anything specific to Digital Services Store - should/could it have a more generic name in case it might be re-used in future for other domains?  Or are there things here that we really only want to use for DSS?

TheDoubleK:
Left a comment but looks good as far as I can tell... :+1:

PR #196, opened 2016-07-04
Switch to AWS Certificate Manager certificate for DM and DSS domains

allait:
screen shot 2016-07-04 at 15 32 54

Setting up a redirect for DSS means we need to have a valid
TLS certificate for the digitalservicesstore.service.gov.uk domain.
Since we're using our existing nginx instances and HTTP listeners
on the nginx ELB we can only set up one certificate per listener -
so we need a single certificate for both digitalmarketplace and DSS
domains. The added benefit of this approach is that we don't have
any SNI issues with old OS and browsers since all requests are covered
by the same certificate.

We've requested a new certificate through ACM, which allows us to
connect it with ELB listeners using our existing template.


TheDoubleK:
:+1:

PR #197, opened 2016-07-05
Add MX and TXT records from DSS DNS to Route53

allait:
Copying DNS records to Route53 so that we don't break Google Mail
when switching the name servers.

### DIFF
--- a/cloudformation_templates/aws_route53_digital_services_store.json
+++ b/cloudformation_templates/aws_route53_digital_services_store.json
@@ -25,6 +25,36 @@
       }
     },
 
+    "GoogleMailServers": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
+        "HostedZoneId": {"Ref": "HostedZone"},
+        "Name": {"Fn::Join": ["", [{"Ref": "Domain"}, "."]]},
+        "Type": "MX",
+        "ResourceRecords": [
+          "10 ALT4.ASPMX.L.GOOGLE.COM.",
+          "10 ALT3.ASPMX.L.GOOGLE.COM.",
+          "5 ALT2.ASPMX.L.GOOGLE.COM.",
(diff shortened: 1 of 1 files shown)


cloudformation_templates/aws_route53_digital_services_store.json
```diff
@@ -25,6 +25,36 @@
       }
     },
 
+    "GoogleMailServers": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
+        "HostedZoneId": {"Ref": "HostedZone"},
+        "Name": {"Fn::Join": ["", [{"Ref": "Domain"}, "."]]},
+        "Type": "MX",
+        "ResourceRecords": [
```
TheDoubleK: Out of interest, any reason these go in reverse order?

cloudformation_templates/aws_route53_digital_services_store.json
```diff
@@ -25,6 +25,36 @@
...
+    },
+
+    "TXTRecords": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
+        "HostedZoneId": {"Ref": "HostedZone"},
+        "Name": {"Fn::Join": ["", [{"Ref": "Domain"}, "."]]},
+        "Type": "TXT",
+        "ResourceRecords": [
+          "\"v=spf1 mx include:_spf.google.com include:amazonses.com ?all\""
```
TheDoubleK: And why `?all` rather than `-all`

TheDoubleK: OK I just read a bit more about it... `?all` seems weaker than we might want?  Or is it because we're not sure who might legitimately be sending digital services store emails?
(My meagre knowledge was just gleaned from https://emailcopilot.com/blog/how-should-i-end-my-spf-record-all/)

TheDoubleK:
Makes sense to stick with current config for sure :+1:

PR #198, opened 2016-07-08
Fix cloudwatch stream filter

allait:
### Fix cloudwatch log streaming filter to not drop exceptions

Previous filter removed logs for health check requests by looking for log events that have either a wrong `request` or wrong `url`, but exception logs don't have any of these fields, so they got filtered out too.

Changing the filter to skip events that:
* Have `request` field and field value is matching health check
  URL - filters out apache-access health check logs
* Have `url` field and field value is matching health check URL -
  filters out application health check logs
* Have `remoteHost` field and field value is `::1` - filters out
  Apache mod_wsgi internal connection logs

### Drop redundant fields from Kibana log events
Removes duplicate `@id` field, `@message` source, account `@owner` and log group fields since they're all the same for all messages.

### DIFF
--- a/lambdas/cloudwatch_logs/main.js
+++ b/lambdas/cloudwatch_logs/main.js
@@ -71,12 +71,7 @@ function transform(payload) {
         ].join('.');
 
         var source = buildSource(logEvent.message);
-        source['@id'] = logEvent.id;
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:

PR #199, opened 2016-07-28
Upgrade PostgreSQL to 9.5

allait:
Now that 9.4 is in production we can start testing 9.5 in preview
and staging.

Version change still requires a new parameter group resource, since
CloudFormation can't change the parameter group version in-place.


TheDoubleK:
:+1: looks good.

PR #201, opened 2016-08-09
Rollback staging database again again

pcraig3:
Yesterday, there was 10 minutes of downtime when I rolled back the staging database.

Checking if I can avoid or mitigate that problem today.


risicle:
What the hell :+1:

PR #202, opened 2016-08-23
Increase kibana instance size

allait:
### Reduce lambda function timeout to 5s
When Kibana server is under heavy load it stops responding to lambda
requests, which makes the function wait until it's interrupted by
a timeout. Higher timeout value means the function will wait longer,
which increases the AWS bill.

Most successfull requests to Elasticsearch should complete within 5s,
so setting that as the timeout value. It could be adjusted if the
lambda monitoring shows a lot of functions being interrupted.

### Increase Kibana isntance size to m3.medium
It looks like micro instances aren't handling the load very well,
both preview and production instances stopped processing requests
after being up for a month. Increasing the instance size to see if
it solves the problem.

Also increasing the disk size. EBS is pretty cheap for the volumes
we're using, and log indexes take up around 400Mb per day in preview
and production.

### DIFF
--- a/cloudformation_templates/aws_kibana.json
+++ b/cloudformation_templates/aws_kibana.json
@@ -6,6 +6,10 @@
       "Type": "String",
       "Description": "Domain name for AWS Elasticsearch service"
     },
+    "InstanceType": {
+      "Type": "String",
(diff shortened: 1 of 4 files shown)


TheDoubleK:
:+1:

PR #204, opened 2016-09-14
DKIM records for Zendesk and Mailchimp

allait:
### Add Zendesk DKIM records
Zendesk uses CNAME records for DKIM authentication and requires two
separate records to be set up.

https://support.zendesk.com/hc/en-us/articles/203663326-Digitally-signing-your-email-with-DKIM-or-DMARC

This might require a change in Zendesk configuration to enable signing

### Add Mailchimp DKIM and SPF records
We're still using Mailchimp for non-transactional emails to buyers
and suppliers, so we need to set up email authentication for it.

http://kb.mailchimp.com/accounts/email-authentication/set-up-custom-domain-authentication-dkim-and-spf

### DIFF
--- a/cloudformation_templates/aws_route53_root_records.json
+++ b/cloudformation_templates/aws_route53_root_records.json
@@ -42,6 +42,21 @@
       }
     },
 
+    "MailchimpDKIMRecord": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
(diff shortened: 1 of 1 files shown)


TheDoubleK:
An extra :+1: because 4th wall isn't picking up the previous one for some reason.

PR #205, opened 2016-09-18
Publish Jenkins releases to Github Deployments API

allait:
### Add a base CLI group to allow commands without stage/app
Not all CLI commands require stack information, so adding a separate `base` click group. Both groups are used for the `dmaws` command.

This drops the auto environment variable support that we're not using. Only explicit variables set for options will be picked up by click.

### Add dmaws.github to publish releases to Github Deployments API
Creates a github deployment for a given repository, ref and environment and sets the deployment status.

### Add a helper command to export Jenkins deployments to GitHub
Requires a JSON export of Jenkins deployments and creates a Github Deployment for each build.

### Publish a Github Deployment after each successful release
Adds a publish_deployment step to the release job. Requires a GITHUB_TOKEN and Jenkins build environment variables to be set.

### DIFF
--- a/dmaws/commands/deploy.py
+++ b/dmaws/commands/deploy.py
@@ -1,10 +1,13 @@
 import sys
+import json
+from datetime import datetime
 
 import click
 
-from ..cli import cli_command
+from ..cli import cli_command, base
 from ..stacks import StackPlan
 from ..build import get_application_name
+from ..github import publish_deployment
 
 
 @click.argument('repository_path', nargs=1, type=click.Path(exists=True))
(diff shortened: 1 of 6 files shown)


dmaws/github.py
```diff
@@ -0,0 +1,83 @@
...
+def publish_deployment(token, repo, ref, environment, build, created_at, ci_url, status=None,
+                       logger=None):
+    log = logger or (lambda *args, **kwargs: None)
+
+    ref_deployments = get_github_deployments(token, repo, ref, environment)
+    if ref_deployments.status_code == 200 and ref_deployments.json():
+        log('Deployment for {}@{} to {} already exists'.format(repo, ref, environment))
+        return True
+
+    created_at = created_at.isoformat()[:-7] + 'Z'
```
TheDoubleK: Feels like we could do with a test around this setting the right value for `created_at`.

TheDoubleK:
Left a couple of minor comments but pretty much looks good.

dmaws/commands/deploy.py
```diff
@@ -24,3 +27,26 @@ def deploy_cmd(ctx, repository_path):
...
+
+@click.argument('deployments_json', nargs=1, type=click.File())
+@click.option('--github-token', help="Github API token", default=None)
+@base.command('publish-deployments')
+def publish_releases(deployments_json, github_token):
+    """Publishes Jenkins deployments JSON dump to Github Deployments API"""
+
+    releases = json.load(deployments_json)
+    return all(
+        publish_deployment(
```
allait: I think we want those keys to exist, otherwise we're not publishing the correct data. That JSON is manually built at the moment and we shouldn't normally need this once each release publishes it's own data.

dmaws/github.py
allait: Added a payload `assert`

TheDoubleK:
:+1:

PR #206, opened 2016-09-20
Move stage argument after the command name

allait:
There doesn't seem to be a way to organize commands with and without arguments before the command name in a single CLI, so in order to support the helper commands I'm moving the STAGE argument after the command name.

This also removes ENVIRONMENT argument. We haven't used environment names that don't match the stage name until now, and we don't have any plans to at the moment, so I'm hiding the separate argument and making sure that environment is always the same as stage.


TheDoubleK:
Can't see anything obviously mad... suck it and see? :+1:

PR #211, opened 2016-09-26
Add Google DKIM record to digtalmarketplace.service.gov.uk DNS

allait:
We're using GMail/Google Groups to send group emails from the support email address, so we need to validate them in order to pass DMARC checks.

The DKIM record is different for each account, so it needs to be generated through Google Apps administrator page.

https://support.google.com/a/answer/173535?hl=en&ref_topic=2752442

### DIFF
--- a/cloudformation_templates/aws_route53_root_records.json
+++ b/cloudformation_templates/aws_route53_root_records.json
@@ -82,6 +82,21 @@
       }
     },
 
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Looks suitably similar to other DKIM records so hopefully should work OK :+1:

PR #212, opened 2016-09-27
Split Google DKIM record to meet TXT length limit

allait:
Route53 limits TXT records to 255 characters, so we need to split Google DKIM record into two quoted strings.

https://support.google.com/a/answer/173535?hl=en&ref_topic=2752442


TheDoubleK:
: 
+ 
1 
: 
:+1:

PR #213, opened 2016-09-28
Fix Google DKIM record split

allait:
Splitting the record in two stops it from being recognized by DKIM checkers. After some investigation it looks like the way to do it is to use 👉 a single space between two quoted strings 👈 instead of a newline. This is accepted by Route53 despite being longer than 255 characters and seems to be recognized by DKIM checking tools.

https://stelfox.net/blog/2014/07/spf-and-dkim-records-in-route-53/


TheDoubleK:
We can only try I guess... :+1:

PR #215, opened 2016-11-18
set/unset some security related headers in nginx

leohemsted:
clear X-Forwarded-Host: if it was set to www.evil.com, and we 302 to /foo,
the browser would redirect to www.evil.com/foo

add X-Content-Type-Options nosniff: stop browsers inferring MIME types, which
is bad and stuff.

set X-Xss-Protection, which is enabled by default in most browsers and
completely stops rendering if XSS is detected

for more information on the headers, have a look at all the useful links in
https://gist.github.com/plentz/6737338

we've already got Strict-Transport-Security enabled everywhere other than the healthcheck conf (https://github.com/alphagov/digitalmarketplace-aws/blob/master/playbooks/roles/nginx/templates/healthcheck.j2)

### DIFF
--- a/playbooks/roles/nginx/templates/api.j2
+++ b/playbooks/roles/nginx/templates/api.j2
@@ -4,8 +4,6 @@ server {
 
     set $api_url "{{ api_url }}";
 
-    add_header Strict-Transport-Security "max-age=31536000; includeSubdomains";
-
     location / {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto https;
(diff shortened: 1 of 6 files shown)


allait:
Should we move the HSTS one to `nginx.conf` too? I think it's repeated in each server config at the moment and it'd be nice to keep the security stuff all together

leohemsted:
provided it won't cause an issue to the healthchecks, which might be http? but i can't imagine it being an issue cos it's only a suggestion to the browser that it should use https instead, rather than a strict command. i'll do it and we can see if everything explodes i guess¯\\\_(ツ)\_/¯

allait:
👍

PR #218, opened 2016-11-21
add dev_user_ips to nginx template

leohemsted:
build was failing cos of it. don't think i need it anywhere else? stacks.yml already refers to `dev_user_ips`

fao @allait


allait:
👍

PR #219, opened 2016-11-22
always set nginx proxy headers

leohemsted:
proxy_set_header directives are only considered at the most deeply nested
level, so remove them all from everywhere and only define them in nginx.conf.j2


allait:
👍

PR #220, opened 2016-11-22
Update nginx AMI

allait:
Fixes proxy_set_header issue with X-Forwarded-Host not being removed from the forwarded headers.


leohemsted:
👌

PR #223, opened 2016-12-09
Add S3 logs bucket and enable S3 access logging

allait:
Enables access logging for all S3 buckets. Logs are stored in a separate logs bucket for each environment and logs from buckets are stored in separate folders. Log bucket access logs are stored in the same bucket (which thankfully doesn't create an infinite feedback loop).

Enabling logging was recommended by ITHC report. I'm not sure what the cost implications are since it seems to create a lot of small files, and S3 could get quite expensive for LIST/PUT operations, so we'll have to keep an eye on it.

### DIFF
--- a/cloudformation_templates/aws_s3.json
+++ b/cloudformation_templates/aws_s3.json
@@ -5,6 +5,15 @@
     "BucketName": {
       "Type": "String",
       "Description": "S3 bucket name"
+    },
+    "AccessControl": {
+      "Type": "String",
+      "Description": "S3 bucket ACL",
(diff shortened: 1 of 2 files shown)


pcraig3:
So I've had a look through the code and I broadly get it, but not very well.
Here's what I think is happening:

- we're logging access (reads and writes) from any of our s3 buckets on all stages
- each stage gets its own bucket for all logs (including its own logs?)
- everything is all wired up once this goes through: no additional config is needed beyond what's here now

pcraig3:
👍

PR #224, opened 2016-12-16
Clean production data for other environments

pcraig3:
The data we collect has changed (grown, mostly) since this data migration logic was last used, and so we aren't removing all the real personal data that we should.  

We delete all draft things because they've never been made public and we try to remove real names and email addresses where we've asked for them:

- Draft briefs are deleted
- All framework agreements are deleted
- All brief responses are deleted
- Two more fields are added to the declaration because we need them for the updated signing flow

### DIFF
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -256,17 +256,14 @@ class RDSPostgresClient(object):
                 "DELETE FROM supplier_frameworks WHERE framework_id IN %s",
                 (open_framework_ids,))
 
+        # We can't tell if a procurement is ongoing, so delete all of them
+        self.log("Delete brief_responses")
+        self.cursor.execute("DELETE FROM brief_responses")
+
         self.log("Delete draft briefs")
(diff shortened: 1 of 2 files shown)


dmaws/rds.py
```diff
@@ -287,59 +294,28 @@ def clean_database_for_staging(self):
         self.log("Delete audit events")
         self.cursor.execute("DELETE FROM audit_events")
 
+        self.log("Delete draft framework agreements")
```
allait: Should we drop all of them? I don't think there's anything special about the draft ones. And the files aren't there anyway, so it's not like they'd be useful for the admin flow.

pcraig3: Yeah, fair enough.

pcraig3:
Gonna merge this soon because I've run it a few times and smoothed out all the kinks.

allait:
Yeah, go for it, it looks good 👍

PR #225, opened 2016-12-20
Rollback staging database to this morning's version

pcraig3:
DOS-2 is closed and we need some preset data to check that the scripts determining who is on the framework do the right thing.


TheDoubleK:
:+1:

PR #226, opened 2016-12-23
Rollback staging database again at the behest of @andrewchong

pcraig3:
Easy as.


pcraig3:
Wow, Paul, this looks great! 👍

PR #227, opened 2017-01-03
Delete database instance if one exists already

pcraig3:
- Start trying to any leftover database instances before creating a new one.  If the script failed before and didn't delete its old database instance, then it would be impossible to run the migration again without manually deleting it.  

- Also upgrades `bcrypt` to get rid of a warning message.

### DIFF
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -152,9 +152,11 @@ class RDS(object):
         if not instance:
             return
 
-        self.log("Found leftover RDS instance: {}:{}".format(
-            instance.id, instance.DBName
-        ))
+        log_msg = "Found leftover RDS instance: {}".format(instance.id)
+        if hasattr(instance, 'DBName'):
+            log_msg += ":{}".format(instance.DBName)
+
+        self.log(log_msg)
         self.delete_instance(instance_id)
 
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Did we ever work out why it sometimes fails to clean up instances in the first place?

This looks like it will do the job, but it might be better to ensure that old ones aren't left about at the end of a run rather than having to tidy up at the start of a subsequent run?

pcraig3:
Yeah, I was thinking that too, but it felt like an odd thing to try and do.  Basically, the failures I've seen have been odd exceptions being thrown at different parts of the process, so it's hard to know where to expect them to come from.

Having said that, maybe we could surround everything in the [migratedata](https://github.com/alphagov/digitalmarketplace-aws/blob/master/dmaws/commands/migratedata.py#L19) method with a big `try/finally:`, where we do all the cleanup.

Do you think that's a better approach, Kev?

TheDoubleK:
I'd like to hear from @allait on this, but to me it feels nicer to tidy up at the end of a run rather than the beginning of a new one.

pcraig3:
Alexey agrees with you, by the way.  But now we're working on saving the database dump somewhere else so that we can give developers access to it.

PR #229, opened 2017-01-06
Pass in correct profile to the StackPlan + open `database_dev_access` for target stage

pcraig3:
####	Set the profile correctly for StackPlan

 The `ctx.stage` value is passed in as a `profile_name` to the `RDS` initialisation, but [it doesn't get passed properly to the `StackPlan`](https://github.com/alphagov/digitalmarketplace-aws/blob/master/dmaws/commands/migratedata.py#L38-L39).  Added an extra bit to the `from_ctx()` method that fixes this.

#### Open `database_dev_access` to target stage

The way I understand it, we've been creating a security group for our new `exportdata` instance (which lets Jenkins connect to the database) and then attaching it to the target stage (where we're going to import the data) as well so that Jenkins can pg_dump to the target database.

Reusing our new security group this way has occasionally meant -- if the script fails to clean up everything -- we leave open access to our target database which then never gets cleaned up.

Opening `database_dev_access` on the target stage gives jenkins access (so the migration script works like it did before) and it also gets closed nightly, so there's less potential for security holes.

### DIFF
--- a/dmaws/commands/migratedata.py
+++ b/dmaws/commands/migratedata.py
@@ -62,24 +62,24 @@ def create_scrubbed_instance(ctx, target_stage):
     return rds, pg_client
 
 
-def dump_to_target(ctx, src_pg_client):
-    rds = RDS(ctx.variables['aws_region'], logger=ctx.log, profile_name=ctx.stage)
-    plan = StackPlan.from_ctx(ctx, apps=['database'])
+def dump_to_target(target_ctx, src_pg_client):
+    rds = RDS(target_ctx.variables['aws_region'], logger=target_ctx.log, profile_name=target_ctx.stage)
+    plan = StackPlan.from_ctx(target_ctx, apps=['database'])
     plan.info()
 
     instance = rds.get_instance(plan.get_value('stacks.database.outputs')['URL'])
 
-    StackPlan.from_ctx(ctx, apps=['database_dev_access'], profile_name=ctx.stage).create()
+    StackPlan.from_ctx(target_ctx, apps=['database_dev_access'], profile_name=target_ctx.stage).create()
 
     pg_client = RDSPostgresClient.from_boto(
         instance,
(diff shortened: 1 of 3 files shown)


dmaws/stacks.py
```diff
@@ -195,7 +198,7 @@ def from_ctx(cls, ctx, **kwargs):
             variables=kwargs.get('variables', ctx.variables),
             apps=kwargs.get('apps', ctx.apps),
             logger=kwargs.get('logger', ctx.log),
-            profile_name=kwargs.get('profile_name'),
+            profile_name=kwargs.get('profile_name', ctx.stage)
```
allait: Setting a default here feels a bit scary, since it could mean that it will start using your production credentials when you don't expect it. Can we pass the profile_name in `dump_to_target` explicitly?

pcraig3: yup.

dmaws/commands/migratedata.py
```diff
@@ -68,10 +69,7 @@ def dump_to_target(ctx, src_pg_client):
 
     instance = rds.get_instance(plan.get_value('stacks.database.outputs')['URL'])
 
-    instance = rds.allow_access_to_instance(
-        instance, IMPORT_SECURITY_GROUP_NAME,
-        ctx.variables['dev_user_ips'],
-        ctx.variables['vpc_id'])
+    StackPlan.from_ctx(ctx, apps=['database_dev_access']).create()
```
allait: I think these are the places you need to pass `profile_name` instead of in `create_scrubbed_instance`.
The original `ctx` will match the profile, since Jenkins takes care of that by setting `AWS_PROFILE` for the run. It's when we switch to the target context that we need to override it with the target stage.

pcraig3: Gotcha.  Updated it.

allait:
👍

PR #230, opened 2017-01-06
Change database names

pcraig3:
- blanking out the snapshot id means that when we run the `create database` job, it will create a new, empty database for us. 

- changing the database name means that new databases will have new names.  this one is pretty self-explanatory.

Taken together, this means that next time we run a `create database` command (for example `dmaws create database staging`) we will create an new empty instance in its place with a new name.
And then we can ruin our beautiful new minimalist database with a bunch of cluttered data using the migrate data job.  Means we'll always have the latest data from production, although it also means we'll have an outage of about 30-40 minutes when we run this.

### DIFF
--- a/vars/preview.yml
+++ b/vars/preview.yml
@@ -9,7 +9,7 @@ database:
   multi_az: "false"
   backup_retention_period: "1"
   slow_query_log_duration_ms: 1000
-  snapshot_id: "s2015-05-13t0953z"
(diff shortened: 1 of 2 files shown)


pcraig3:
Should I close the pull request then or leave it here until it's ready?

pcraig3:
Gonna close this until needed to try and unclog the 4th wall thing.

PR #231, opened 2017-01-11
Delete draft services for open frameworks

pcraig3:
Once a framework is no longer open, we're allowed to see the draft services that were submitted.

<sup>(small change)</sup>


allait:
👍

PR #233, opened 2017-01-17
Delete less suppliers

pcraig3:
We have two queries in a row which seem related to each other:

- one query adds supplier frameworks to suppliers who have services (but no supplier frameworks)
- a second query deletes suppliers without supplier frameworks

It looks like the first query exists as setup for the second query.

This pull request makes the second query more specific so that we can remove the first query entirely.

Also, added some more comments that hopefully make it easier to see why we're changing what we're changing.

### DIFF
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -249,6 +249,7 @@ class RDSPostgresClient(object):
 
         open_framework_ids = self.get_open_framework_ids()
 
+        # Remove data about currently ongoing framework applications
         self.log("Delete draft services for open frameworks")
         self.cursor.execute("""
             DELETE FROM draft_services WHERE framework_id IN (
@@ -262,8 +263,9 @@ class RDSPostgresClient(object):
                 "DELETE FROM supplier_frameworks WHERE framework_id IN %s",
                 (open_framework_ids,))
 
-        # We can't tell if a procurement is ongoing, so delete all of them
+        # Remove data related to open DOS procurements
         self.log("Delete brief_responses")
+        # (We can't tell if a procurement is ongoing, so delete all brief responses)
         self.cursor.execute("DELETE FROM brief_responses")
 
         self.log("Delete draft briefs")
@@ -276,6 +278,7 @@ class RDSPostgresClient(object):
         self.log("  > Delete draft briefs")
         self.cursor.execute("DELETE FROM briefs WHERE published_at IS NULL")
 
+        # Remove suppliers without framework agreements or submitted services
         self.log("Delete dangling suppliers")
         self.cursor.execute("""
             WITH dangling_suppliers AS (
@@ -295,12 +298,16 @@ class RDSPostgresClient(object):
             DELETE FROM suppliers WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
         """)
 
+        # Remove audit events because we don't use them
         self.log("Delete audit events")
         self.cursor.execute("DELETE FROM audit_events")
 
+        # Remove framework agreements because they contain personal data and we don't rely on them
         self.log("Delete framework agreements")
         self.cursor.execute("DELETE FROM framework_agreements")
 
+        # Overwrite declarations with the smallest possible valid entry
+        # Removes all personal data while keeping our app working as expected
         self.log("Blank out declarations")
         self.cursor.execute("""
             UPDATE supplier_frameworks
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -248,6 +248,7 @@ class RDSPostgresClient(object):
         )
 
         open_framework_ids = self.get_open_framework_ids()
+
         self.log("Delete draft services for open frameworks")
         self.cursor.execute("""
             DELETE FROM draft_services WHERE framework_id IN (
@@ -275,18 +276,6 @@ class RDSPostgresClient(object):
         self.log("  > Delete draft briefs")
         self.cursor.execute("DELETE FROM briefs WHERE published_at IS NULL")
 
-        # Fix suppliers with services but no supplier_framework
-        self.cursor.execute("""
-            INSERT INTO supplier_frameworks (supplier_id, framework_id) (
-                SELECT DISTINCT suppliers.supplier_id, services.framework_id
-                FROM suppliers
-                JOIN services ON suppliers.supplier_id=services.supplier_id
-                WHERE (
-                    SELECT COUNT(*) FROM supplier_frameworks WHERE supplier_id=suppliers.supplier_id
-                ) = 0
-            )
-        """)
-
         self.log("Delete dangling suppliers")
         self.cursor.execute("""
             WITH dangling_suppliers AS (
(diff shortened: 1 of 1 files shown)


dmaws/rds.py
```diff
@@ -248,13 +248,24 @@ def clean_database(self):
...
+        self.log("  > Delete draft services for open frameworks")
         self.cursor.execute("""
             DELETE FROM draft_services WHERE framework_id IN (
                 SELECT id FROM frameworks WHERE status='open'
             )
             """)
 
+        self.log("  > Delete draft services belonging to dangling suppliers")
+        self.cursor.execute("""
+            DELETE FROM draft_services WHERE supplier_id IN (
```
allait: I think this could be written as 
```sql
DELETE FROM draft_services WHERE supplier_id NOT IN (
    SELECT supplier_id FROM suppliers INNER JOIN supplier_frameworks on suppliers.supplier_id=supplier_frameworks.supplier_id
)
```

Not sure if it's batter. Or whether this matters at all.

pcraig3: Your suggestion works but I think we should keep what I had before because that particular snippet of SQL shows up a few times in this file.

pcraig3:
Also, I've done an audit into what we start with and what this stuff is doing to it.

Initially, we have

```
7891 suppliers in total
1631 suppliers with no supplier_framework
 136 suppliers that have services but no supplier_framework (all G4 services)
   1 supplier with a draft service but no supplier framework (on G7)
```

***

And this is what happens when we run everything.

```sql
-- Delete draft services for open frameworks
DELETE FROM draft_services WHERE framework_id IN (SELECT id FROM frameworks WHERE status='open')
```
DELETE 0

```sql
-- Delete draft services belonging to dangling suppliers
DELETE FROM draft_services WHERE supplier_id IN (SELECT supplier_id FROM suppliers WHERE (SELECT COUNT(*) FROM supplier_frameworks WHERE supplier_id=suppliers.supplier_id) = 0)
```
DELETE 1

```sql
-- Delete supplier frameworks for open frameworks
DELETE FROM supplier_frameworks WHERE framework_id IN {open_framework_ids};
```
DELETE 0

```sql
-- Delete brief_responses
DELETE FROM brief_responses;
```
DELETE 12892

```sql
-- Delete draft brief users
DELETE FROM brief_users WHERE brief_id IN (SELECT id FROM briefs WHERE published_at IS NULL)
```
DELETE 710

```sql
-- Delete draft briefs
DELETE FROM briefs WHERE published_at IS NULL
```
DELETE 710

```sql
-- Fix suppliers with services but no supplier_framework
INSERT INTO supplier_frameworks (supplier_id, framework_id) ( SELECT DISTINCT suppliers.supplier_id, services.framework_id FROM suppliers JOIN services ON suppliers.supplier_id=services.supplier_id WHERE ( SELECT COUNT(*) FROM supplier_frameworks WHERE supplier_id=suppliers.supplier_id ) = 0 )
```
INSERT 0 136
(I don't know what the 0 is, but we have 136 suppliers with services but no supplier_framework, so this seems correct)

```sql
-- Delete dangling suppliers (Suppliers that are not connected to any frameworks)
SELECT supplier_id FROM suppliers WHERE ( SELECT COUNT(*) FROM supplier_frameworks WHERE supplier_id=suppliers.supplier_id ) = 0 ), d1 AS ( DELETE FROM contact_information WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers) ), d2 AS ( DELETE FROM users WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers) ) DELETE FROM suppliers WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
```
DELETE 1495 (ie, 1631 - 136 = 1495)

```sql
-- Delete audit events
DELETE FROM audit_events
```
DELETE 1054512

```sql
-- Delete framework agreements
DELETE FROM framework_agreements
```
DELETE 5317

pcraig3:
I've updated this pull request based on the work done on friday (deleting the rogue draft service) and I think it's merge-able now.

dmaws/rds.py
```diff
@@ -295,20 +287,27 @@ def clean_database(self):
...
                 ), d1 AS (
                     DELETE FROM contact_information WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
                 ), d2 AS (
                     DELETE FROM users WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
                 )
-             DELETE FROM suppliers WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
+            DELETE FROM suppliers WHERE supplier_id IN (SELECT supplier_id FROM dangling_suppliers)
         """)
 
+        # Remove audit events because we don't use them
```
TheDoubleK: Some audit events are used... Admin uses service update events to show diffs (upcoming work on this v.soon from submissions). And some others (draft service updates maybe?) Used to find dates to display.

I'm not saying we should keep these necessarily, but it's wrong to say that we don't use them at all.

dmaws/rds.py
```diff
@@ -295,20 +287,27 @@ def clean_database(self):
...
         self.log("Delete audit events")
         self.cursor.execute("DELETE FROM audit_events")
 
+        # Remove framework agreements because they contain personal data and we don't rely on them
         self.log("Delete framework agreements")
         self.cursor.execute("DELETE FROM framework_agreements")
 
+        # Overwrite declarations with the smallest possible valid entry
```
TheDoubleK: We use declaration fields for a lot of the close-of-framework scripts, and also in the admin front-end ("nameOfOrganisation", company number and address details are rendered in the countersigning flow.

dmaws/rds.py
pcraig3: Yes, I remember we talked about this the other day.

Probably worth keeping them in then, eh?

dmaws/rds.py
allait: Let's add the keys that are used by the apps with sanitized values. I don't think we can keep the entire declaration

TheDoubleK: That would be good if possible I think.  The bits to look at for required keys are the signing templates in supplier app and countersigning in admin app.  

I think it's namOfOrganisation, the various address fields (three for G8 and DOS2, one for earlier frameworks) and company number.  But there might be a couple more that get referenced.

dmaws/rds.py
TheDoubleK: I'm not sure... if we have no "changed" audit events we should still be able to test the service editing flow - it will should just treat the current version of services as fresh and new.  Feels like a pain to keep just certain events.  And the draft edit date ones I think only make sense during a live procurement, so those drafts shouldn't be copied over anyway.  

I just remember seeing some pages break when those audit events have been deleted.

pcraig3: After talking about this in person yesterday, we agreed not to preserve any audit events (this is what we're doing now).  We could cherry pick some to help us with given scenarios, but ideally we should just be able to create the audit events we need for our test scenarios. (for example, if we're testing the diff page, it's quite easy to make a simple edit to a service and in that way generate an audit event.)

dmaws/rds.py
pcraig3: I've kept the current declaration and run through a few scenarios to see what breaks. 

1. signing an agreement as a supplier
2. putting an agreement on hold as an admin
3. countersigning an agreement as an admin

Turns out that nothing breaks.  We get some empty fields (including an empty link(!) for the companies' house address), but that doesn't actually cause an error.

This is the result: 
screen shot 2017-01-25 at 15 45 51

Personally, I'd argue that keeping the declaration as small as possible is more important than a fake address and a bad link, so if we're happy with this then I'll just merge this PR as-is.

TheDoubleK:
Do it!  :+1:

PR #234, opened 2017-01-17
[#137354029] Enable CloudTrail logging to S3 and CloudWatch

allait:
Sets up CloudTrail to store logs in S3 bucket and creates a separate CloudWatch log group to send logs to.

gds-audit-cloudtrails S3 bucket is controlled by GDS Security Team, and they've asked us to use AWS account ID as key prefix.

CloudWatch is setup in the same AWS account as the CloudTrail, so we have access to view the history of events and can create CloudWatch alerts in the future.

There's only one CloudTrail per AWS account, so the stack name doesn't include stage or environment names (similar to Route53 stacks).

IAM policies and CloudTrail settings are taken from https://github.com/alphagov/aws-security-alarms/blob/master/terraform/main.tf

From AWS docs:

> AWS CloudTrail is a web service that records AWS API calls for your
> account and delivers log files to you. The recorded information includes
> the identity of the API caller, the time of the API call, the source IP
> address of the API caller, the request parameters, and the response
> elements returned by the AWS service.

> With CloudTrail, you can get a history of AWS API calls for your
> account, including API calls made via the AWS Management Console, AWS
> SDKs, command line tools, and higher-level AWS services (such as AWS
> CloudFormation). The AWS API call history produced by CloudTrail enables
> security analysis, resource change tracking, and compliance auditing.

### DIFF
--- a/cloudformation_templates/aws_cloudtrail.json
+++ b/cloudformation_templates/aws_cloudtrail.json
@@ -1,0 +1,67 @@
+{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+  "Parameters" : {
+    "S3BucketName": {
+      "Description": "S3 Bucket to store CloudTrail logs",
+      "Type": "String"
+    }
+  },
+  "Resources" : {
+    "LogGroup": {
+      "Type": "AWS::Logs::LogGroup",
+      "Properties": {
+        "LogGroupName": "cloudtrail",
+        "RetentionInDays": 90
+      }
+    },
+
+    "IAMRole": {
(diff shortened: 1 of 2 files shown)


pcraig3:
Can I thumbs-up this and then see the logs once it's all working?

allait:
@pcraig3 I've run it on preview when I was writing it since it's the only way to test it, so you can take a look at the CloudTrail and the `cloudtrail` CloudWatch log group there

pcraig3:
Oh wow, all the stuff I do is in there :/

👍

PR #235, opened 2017-01-20
Dump exportdata into a local directory

pcraig3:
Instead of streaming our `pg_dump` output straight into the target database, this code creates an intermediary `exportdata.sql` file in a particular directory.

The (relative) filepath is a new parameter passed into the `migratedata` command, and then the python code will create the directory structure and dump the .sql file into it before importing it into our target database.
We're relying on [the migratadata `-jenkins` job](https://github.gds/gds/digitalmarketplace-jenkins/pull/73) to do the sync, and then do the cleanup.

We can decide either to do or not do the 'export to a file' part of this based on whether or not we pass in a filepath. Tested locally and both work.


pcraig3:
💥

PR #236, opened 2017-01-26
[preview] Add terraform for IAM management

bandesz:
This PR introduces Terraform modules for IAM management (and possibly for other AWS resources later.)

The Terraform project is placed in the terraform directory, you can find instructions for setup and usage in the README.md file.

## IAM changes

  - all IAM users will be only created in the digitalmarketplace accounts and everyone will have the proper switch role permissions. (E.g. admins can go to dev/prod accounts as admins, devs can got to (selectively) dev/prod accounts as devs, etc.)
  - in the AWS console you can easily switch user with a single click and locally you can set up the role switch in ~/.aws/config (there will be more info about this in the dm-docs repo)  
  - there will be a stronger password policy in effect: min. 12 characters, and min. one lowercase/uppercase/number each (the PR needs to be updated for this)

### New policies
 - IP restrictions: every IAM user and role is IP restricted to office IPs (and Jenkins), so if you need to work from outside the office you'll need VPN in the future
 - MFA restrictions: every console access rule will enforce MFA authentication (but current and new users will be still able to set it up anytime)

### New groups
  - Admins: administrators with MFA restrictions - don't use access keys for these users unless you want to handle MFA from cli
  - Developers: read/only for the necessary services, S3 dev access and 
  - Terraform: API access for running Terraform
  - Ansible: API access for running Ansible
  - Packer: API access for running Packer
  - Basic Users: only login access
  - SwitchToDevDeveloper: ability to switch to the developer role in the dev account
  - SwitchToProdDeveloper: ability to switch to the developer role in the prod account
  - SwitchToDevS3Only: ability to switch to the s3 only role in the dev account (can be used together with the Basic User group)

## Secret management

The account ids and the group memberships are stored in the creds repo with SOPS. With the Terraform AWS credentials you can automatically assume a proper role to decrypt the SOPS credentials. More info in the related PR: https://github.gds/gds/digitalmarketplace-credentials/pull/94

The PR also creates two KMS master keys (in two different regions) to be used for SOPS.

## TODO
 - [x] add all users
 - [ ] finalise password policy (postponed)


terraform/modules/sops-credentials/kms.tf
```diff
@@ -0,0 +1,71 @@
...
+}
+EOF
+}
+
+resource "aws_kms_alias" "sops_credentials_primary" {
+  name = "alias/sops-credentials"
+  target_key_id = "${aws_kms_key.sops_credentials_primary.key_id}"
+}
+
+resource "aws_kms_key" "sops_credentials_secondary" {
```
allait: Why do we need another key?

terraform/modules/iam-common/developers.tf
```diff
@@ -0,0 +1,95 @@
+resource "aws_iam_policy" "developer" {
+  name = "Developer"
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "sqs:Get*",
```
bandesz: Yeah, I'll sort them.

terraform/modules/iam-common/developers.tf
```diff
@@ -0,0 +1,95 @@
...
+      "Effect": "Allow",
+      "Resource": "*",
+      "Condition": {
+        "Bool": {
+          "aws:MultiFactorAuthPresent": true
+        }
+      }
+    },
+    {
+      "Action": [
```
bandesz: It was a copy-paste originally, I moved these to the big list as "s3:Get*" and "s3:List*"

terraform/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The remote state files are stored on S3 and the initialisation is automatic when you run the terraform make targets (see below).
+
+### Local secret storage
+
+Follow the set up guide outlined here: https://github.gds/gds/digitalmarketplace-credentials/blob/master/README.md
+
+The Makefile will look for the credentials repository at the directory defined in the DM_CREDENTIALS_REPO environment variable.
+
+## Make targets
```
bandesz: I wouldn't put that there as the main README can grow quite big and containing too much information. We can link the Terraform docs in the main README.

terraform/modules/sops-credentials/kms.tf
bandesz: If the Ireland region becomes inaccessible we can still decrypt the credentials with the other master key.

PR #238, opened 2017-02-06
[preview] Add paas-manifest command to dmaws for generating PaaS manifest files

bandesz:
I added a new command to dmaws to generate PaaS manifest files from a Jinja2 template. The command is using the built-in variable loading mechanism.

Subcommand: paas-manifest
Extra options:
```
  -t, --template PATH             Manifest Jinja2 template file
  -c, --credentials-repo PATH     Path to the digitalmarketplace-credentials
                                  repository
  -o, --out-file TEXT             Output file, if empty the template content
                                  is printed to the stdout
```

Components:
 - Paas manifest Jinja2 template: paas/manifest.j2. Can be changed with the --template option.
 - The PaaS configs are added to the existing apps under the paas key (see vars/common.yml)
 - Application secrets are optional and loaded the same way but from the credentials repo (vars/common.yaml and vars/[environment].yaml)

The configurations are merged in the following order:
 - vars/common.ym
 - vars/[environment].yml
 - [credentials repo]/vars/common.yaml
 - [credentials repo]/vars/[environment].yaml

Example usage:
```
dmaws paas-manifest development supplier-frontend
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ paas-deploy: ## Deploys the app to PaaS
 .PHONY: paas-rollback
 paas-rollback: ## Rollbacks the app to the previous release on PaaS
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
+	@[ $$(cf curl /v2/apps/`cf app --guid ${APPLICATION_NAME}-rollback` | jq -r ".entity.state") = "STARTED" ] || (echo "Error: rollback is not possible because ${APPLICATION_NAME}-rollback is not in a started state" && exit 1)
 	cd ${DEPLOYMENT_DIR} && \
 		cf app --guid notify-admin-rollback && \
 		cf delete -f notify-admin && \
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,3 @@ paas-push: ## Pushes the app to PaaS
 .PHONY: paas-clean
 paas-clean: ## Cleans up all files created for the PaaS deployment
 	rm -rf ${DEPLOYMENT_DIR}
-
-build-app:
-	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,87 @@
+.DEFAULT_GOAL := help
+SHELL := /bin/bash
+VIRTUALENV_ROOT := $(shell [ -z ${VIRTUAL_ENV} ] && echo $$(pwd)/venv || echo ${VIRTUAL_ENV})
+
+PAAS_API ?= api.cloud.service.gov.uk
+PAAS_ORG ?= digitalmarketplace
+PAAS_SPACE ?= ${STAGE}
+
+DEPLOYMENT_DIR := ${CURDIR}/tmp
+CF_HOME ?= ${DEPLOYMENT_DIR}
+$(eval export CF_HOME)
+
+.PHONY: help
+help:
+	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: requirements
+requirements: ## Install requirements
+	${VIRTUALENV_ROOT}/bin/pip install -e .
+
+.PHONY: virtualenv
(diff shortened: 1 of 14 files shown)


dmaws/commands/paasmanifest.py
```diff
@@ -0,0 +1,51 @@
...
+    manifest_content = template_string(templace_content, variables)
+
+    if out_file is not None:
+        with open(out_file, 'w') as f:
+            f.write(manifest_content)
+        os.chmod(out_file, 0o600)
+    else:
+        print(manifest_content)
+
+def get_secret_vars(credentials_repo, filename):
```
allait: I think we should pass in a decrypted file with `<(sops_wrapper ...)` and then we won't need this. I think we've tried this and it seemed to work before and it would decouple the manifest generation from the credentials encryption.

bandesz: That's true, I'm surprised why I haven't done that this way in the first place :) Changed.

Makefile
```diff
@@ -0,0 +1,103 @@
...
+.PHONY: paas-login
+paas-login: ## Log in to PaaS
+	$(if ${PAAS_USERNAME},,$(error Must specify PAAS_USERNAME))
+	$(if ${PAAS_PASSWORD},,$(error Must specify PAAS_PASSWORD))
+	$(if ${PAAS_SPACE},,$(error Must specify PAAS_SPACE))
+	mkdir -p ${CF_HOME}
+	@cf login -a "${PAAS_API}" -u ${PAAS_USERNAME} -p "${PAAS_PASSWORD}" -o "${PAAS_ORG}" -s "${PAAS_SPACE}"
+
+.PHONY: paas-deploy
+paas-deploy: ## Deploys the app to PaaS
```
allait: At which point does this start sending live traffic to the new deployment? It doesn't look like it's modifying the routes, so I'm assuming it happens when we change the app name? Or will both apps be receiving a share of the traffic since the have the same routes set?

Makefile
```diff
@@ -0,0 +1,103 @@
...
+	cd ${DEPLOYMENT_DIR} && \
+		cf app --guid ${APPLICATION_NAME} && \
+		cf rename ${APPLICATION_NAME} ${APPLICATION_NAME}-rollback && \
+		cf push -f manifest.yml && \
+		cf scale -i $$(cf curl /v2/apps/$$(cf app --guid ${APPLICATION_NAME}) | jq -r ".entity.instances" && \ 2>/dev/null || echo "1") ${APPLICATION_NAME} && \
+		cf stop ${APPLICATION_NAME}-rollback && \
+		cf delete -f ${APPLICATION_NAME}-rollback
+
+.PHONY: paas-rollback
+paas-rollback: ## Rollbacks the app to the previous release on PaaS
```
allait: What happens if the rollback app has already been stopped, but not deleted?

Makefile
```diff
@@ -0,0 +1,103 @@
...
+paas-push: ## Pushes the app to PaaS
+	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
+	cd ${DEPLOYMENT_DIR} && \
+		cf push ${APPLICATION_NAME} -f manifest.yml
+
+.PHONY: paas-clean
+paas-clean: ## Cleans up all files created for the PaaS deployment
+	rm -rf ${DEPLOYMENT_DIR}
+
+build-app:
```
allait: This doesn't seem to do anything

paas/manifest.j2
```diff
@@ -0,0 +1,31 @@
+---
+
+applications:
+  - name: digitalmarketplace-{{ app|replace('_', '-') }}
+    stack: cflinuxfs2
+    buildpack: python_buildpack
+    command: ./run.sh gunicorn -w 4 -b 0.0.0.0:$PORT application:application --log-file -
```
allait: How does run.sh end up in the application release?

Makefile
bandesz: When the cf push finished the new app version starts getting traffic right away. Then we upscale it to the same number and remove the old version. (It should be only a short amount of time)

Makefile
bandesz: Rollback needs to be always a careful manual process, so you have to check what's the current status. The paas-rollback is just a convenient command for the most common cases.

paas/manifest.j2
bandesz: The Jenkins job adds it when deploying.

Makefile
bandesz: Removed

PR #250, opened 2017-03-17
Remove sops key access from dev and prod users

allait:
Previous IAM role allowed dev and prod account users to assume the
role for accessing sops KMS keys. But our applications never
decrypt the credentials themselves and we're moving all users to
the main account, so we don't need these permissions.


allait:
So we need these permissions because we call sops while switched to the infrastructure roles, and without these permissions we'd have to ask for MFA twice. The problem with this though is that if the dev/production accounts are compromised it's possible for the attacker to create permanent access key to decrypt credentials.

Closing this for now until we think of a better approach.

PR #254, opened 2017-03-29
Fix AWS profile definition in the readme

bandesz:


TheDoubleK:
This is right.  Merging.

PR #256, opened 2017-04-18
Fix bug in `clean_database` script

Wynndow:
When we migrate data from production to another environment, sensitive
data is removed. This includes data about currently ongoing framework
applications (for frameworks that are `open`), such as draft services
and `supplier_frameworks` info.

Previously we were deleting `supplier_frameworks` data for frameworks
that were `pending` as well as `open`.

We were then removing suppliers who had no entry in the
`supplier_frameworks` table as well as no services (dangling suppliers).

Because `supplier_frameworks` info had been deleted for `pending` and
`open` frameworks but draft services only deleted for `open` frameworks,
this led to a foreign key constraint error on the `draft_services` table
when trying to delete suppliers with no `supplier_framework`.

This commit makes the function only delete `supplier_frameworks` for
`open` frameworks.

### DIFF
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -232,10 +232,6 @@ class RDSPostgresClient(object):
             ['psql', '-d', target_pg_client.db_path],
         ], logger=self.log, ignore_errors=False)
 
-    def get_open_framework_ids(self):
-        self.cursor.execute("SELECT id FROM frameworks WHERE status IN ('open', 'pending')")
-        return tuple(framework[0] for framework in self.cursor.fetchall())
-
     def clean_database(self):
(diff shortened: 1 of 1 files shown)


pcraig3:
God, this is so sloppy of @robyoung to have left this in this state.

👍

robyoung:
Hey! What about the [prime directive](https://martinfowler.com/bliki/PrimingPrimeDirective.html) @pcraig3?

pcraig3:
Shhh, @robyoung I'm just trying to deflect attention from my oversight. :sweat_smile:

Anyone who follows the directive won't blame either of us, but those who don't will know it's not my fault.

PR #258, opened 2017-04-20
Add terraform/ansible/nginx setup to enable maintenance mode

Wynndow:
Doesn't seem to have a Pivotal Story, but this is connected [https://www.pivotaltracker.com/story/show/143486469](https://www.pivotaltracker.com/story/show/143486469).

We don't currently have a way to switch the site to a maintenance mode if we have planned down time.

If the `nginx_config` variable is set to `live` in Terraform for a particular environment the site will behave as usual. If it is set to `maintenance` then Ansible will update the Nginx autoscaling group and only load up the maintenance and healthcheck nginx configs.

The healthcheck config allows the ELB healthchecks to pass, and the maintenance config returns a static html file for any url on the www subdomain. This html file needs updating once content and design have looked at it.

### DIFF
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -1,14 +1,14 @@
 ---
-live_config:
-  - www
-  - api
-  - assets
-  - healthcheck
-  - elasticsearch
-  - digitalservicesstore
-
-maintenance_config:
-  - healthcheck
-  - maintenance
+nginx_configs:
+  live:
+    - www
+    - api
+    - assets
+    - healthcheck
+    - elasticsearch
+    - digitalservicesstore
+  maintenance:
+    - healthcheck
+    - maintenance
 
 static_files_root: /usr/share/nginx/html
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -7,6 +7,8 @@ live_config:
   - elasticsearch
   - digitalservicesstore
 
-maintenance_config: maintenance
+maintenance_config:
+  - healthcheck
+  - maintenance
 
 static_files_root: /usr/share/nginx/html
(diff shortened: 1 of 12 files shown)


playbooks/roles/nginx/tasks/main.yml
```diff
@@ -62,21 +66,41 @@
...
+  tags: instance-config
+  when: nginx_config == "live"
+
+- name: Create the Nginx maintenance configuration files
   template: src={{ item }}.j2
             dest=/etc/nginx/sites-available/{{ item }}
             backup=yes
   notify: reload nginx
-  with_items: nginx_configs
+  with_items: "{{ maintenance_config }}"
```
allait: `{{ nginx_configs[current_config] }}` and then we might be able to get rid of the duplicate task

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -1,10 +1,14 @@
 ---
-nginx_configs:
```
allait:
```
nginx_configs:
  live:
    - www
    - ...
  maintenance:
    - maintenance
```

terraform/modules/nginx/autoscaling.tf
```diff
@@ -87,7 +87,8 @@ cd /home/ubuntu/provisioning && ansible-playbook -c local -i localhost, nginx_pl
     -e elasticsearch_auth='${var.elasticsearch_auth}' \
     -e app_auth='${var.app_auth}' \
     -e aws_region='${data.aws_region.current.name}' \
-    -e nameserver_ip="$(awk '/nameserver/{ print $2; exit}' /etc/resolv.conf)"
+    -e nameserver_ip="$(awk '/nameserver/{ print $2; exit}' /etc/resolv.conf)" \
+    -e nginx_config='${var.nginx_config}'
```
allait: I'd call this something like `mode=live` or `mode=maintenance`. As long as it doesn't clash with any existing variables.

PR #261, opened 2017-04-27
Change paas-deploy process to not require manual rollback

allait:
### Change paas-deploy process to not require manual rollback

Previous release process was renaming the app to `{app}-rollback`. If
the cf push failed to bring up the new application version then the
release failed, leaving the rollback application, which means that
on the next release attempt the renaming fails since `{app}-rollback`
is already taken.

The new process instead pushes the new app as `{app}-release`. If
the release succeeds then the old application is unmapped from the
route and deleted and `{app}-release` is renamed to `{app}`.

This change also means that if an app is accidentally pushed using
`cf push` and the generated manifest it won't bring down the production
application during release, but will create a separate `{app}-release`
application instead.

If the release fails then both `{app}` and `{app}-release` will exist,
and the next release will push to `{app}-release` as usual.

This also fixes the need to manually create an app before attempting
the release for the first time.

We're removing the scaling step from the release process since
autoscaling is not set up on PaaS yet so the manifests should serve
as the main source of truth for the desired instance counts.

This also introduces a delay between removing the traffic from old
application version and shutting down to allow the instances to push
any outstanding logs to CloudWatch.

### Set up HTTP healtcheck on /_status for all PaaS apps

By default CloudFoundy would use port healthchecks and will mark the
application as healthy as soon as it starts listening on the specified
port.

Since we already have a /_status endpoint for the apps we can use that
to check that the app is running instead.


Makefile
```diff
@@ -132,4 +132,4 @@ paas-clean: ## Cleans up all files created for the PaaS deployment
 paas-populate-db: ## Imports postgres dump specified with `DB_DUMP=` to targeted spaces db
 	$(call check_space)
 	$(if ${DB_DUMP},,$(error Must specify DB_DUMP))
-	./scripts/paas_populate_db.sh ${DB_DUMP}
+	./scripts/populate-paas-db.sh ${DB_DUMP}
```
Wynndow: Booooo

Wynndow: Booo

PR #263, opened 2017-05-02
🏀  Baller maintenance page 🏀 

pcraig3:
Adding a maintenance page template that looks like [our 404 page](https://www.digitalmarketplace.service.gov.uk/sponge-avengers).

Started with the 404 page HTML and then made a bunch of small changes.  [Second commit](https://github.com/alphagov/digitalmarketplace-aws/commit/7ba21c061563f4a42a2371d62000f4a71429d38c) is probably the most interesting one. Also did some magic behind the scenes to crunch down all that CSS, but it's still gonna weigh in at ~2000 lines.  (aside: our reset CSS rules are repeated 8(!) times in our compiled `application.css`. 😱 ) 

@tracy-content's had a look at this, so I've removed the `WIP`.

## current screenshot 

screen shot 2017-05-08 at 14 33 00

#### Note 

I've created a repository that does all of the CSS munging.  If you want to update the maintenance page or create more static pages with our Digital Marketplace styles, it's a good place to start.  Find it here: https://github.com/alphagov/digitalmarketplace-maintenance

### DIFF
--- a/playbooks/roles/nginx/files/maintenance.html
+++ b/playbooks/roles/nginx/files/maintenance.html
@@ -16,7 +16,500 @@
   <meta content="https://assets.publishing.service.gov.uk/static/opengraph-image.png" property="og:image">
   <!--[if lt IE 9]>
   <script>
-    /* inline JavaScript here for < IE9 */
+    /**
+      * @preserve HTML5 Shiv 3.7.3 | @afarkas @jdalton @jon_neal @rem | MIT/GPL2 Licensed
+    */
+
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Is there a separate page for when we take things down for a long time?

pcraig3:
Nope. This is it, Kev.

galund:
Could there be a script which generates this based on some template plus the shared CSS? Seems like it would be nice if this could update automatically if we change our styles.

TheDoubleK:
Should we drop the word "short" then? Or are we optimistic about being able to recover quickly from any unforseen disaster?

pcraig3:
@galund, I have a local repo that automates some of the process of getting the styles into the page but not all of it. Happy to share it more widely.

Specifically, the steps are:
1. Copy + paste the compiled CSS from the buyer app 
2. Run a script to cut down all the CSS rules 
3. Copy + paste the compiled CSS into the  page`<head>`

Changing the words and the links to gov.uk assets is all manual.

*** 

As to your specific point, my feeling is that this page probably will rarely change because it will even less rarely be used. I would assume it's probably less time to do the manual steps once in a blue moon than to write more code to automate creating more of this page.

galund:
Yeah, if you've got code, then add a note to the readme file here, and put some more detail into the readme for your new repo? Seems like a worthwhile thing to have exist, even if it's only a partial solution and some of the steps are manual.

pcraig3:
@galund, Hey, I've pushed up my repo. Look for it here: https://github.com/alphagov/digitalmarketplace-maintenance

@TheDoubleK, Hey Kev, Alexey says that the "shortly" is fine because we're not expecting to have this page up for more than a day at any point.

playbooks/roles/nginx/files/maintenance.html
```diff
@@ -1 +1,1868 @@
...
+  <div id="wrapper">
+    <main id="content" role="main">
+      <div class="error-page">
+        <div class="index-page grid-row">
+          <div class="column-two-thirds">
+            <header class="page-heading-smaller">
+              <h1>Planned maintenance</h1>
+            </header>
+            <p>The Digital Marketplace is unavailable because of planned maintenance. It’ll be available again soon.</p>
+            <p>If you need help urgently, email<a href="mailto:enquiries@digitalmarketplace.service.gov.uk?subject=Digital%20Marketplace%20feedback" title="Please send feedback to enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a>.</p>
```
carolinegreen: There's no space here and the formatting looks kind of strange.

carolinegreen:
Think you could do with adding a more current screenshot with the new content.

playbooks/roles/nginx/files/maintenance.html
pcraig3: What do you mean?  I don't think there's a problem with the HTML itself.

carolinegreen: screen shot 2017-05-08 at 15 15 50

pcraig3:
[Maintenance page html file](https://trello.com/c/rEykWSOo/58-maintenance-page-html-file)

carolinegreen:
I think this is probably ready to go 👍

PR #269, opened 2017-05-09
Create terraform prod environment

Wynndow:


terraform/environments/production/main.tf
```diff
@@ -0,0 +1,89 @@
...
+      "subnet-ad0894c8"
+  ]
+
+  instance_count = "2"
+  min_instance_count = "2"
+  max_instance_count = "2"
+
+  instance_type = "t2.micro"
+
+  log_retention_days = "180"
```
allait: This should be 10 years for production

PR #270, opened 2017-05-09
Set up log groups for PaaS apps

Wynndow:
Trello: https://trello.com/c/FtoKj86m

When the apps are running in the PaaS, they'll produce application logs and nginx logs. We need to capture these in CloudWatch in separate log groups to aide in processing them.

This sets up the log groups in Terraform. Prod logs need to be kept for at least 10 years.


Wynndow:
[Configuring PaaS environments](https://trello.com/c/FtoKj86m/56-configuring-paas-environments)

PR #273, opened 2017-05-12
Change production domain to digitalmarketplace.service.gov.uk

Wynndow:
The production state file has been updated in S3 so that the nginx route53 zone being managed by Terraform is now the original digitalmarketplace.service.gov.uk zone. This is so that when we make the switch from production.marketplace.team, we don't need to recreate the zone file and lose the name server records.

The domain is updated here to match the new zone file. The original subdomain records for api, assets etc will need to be deleted before this is applied so that Terraform can recreate them.

### DIFF
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -16,7 +16,7 @@ module "production_nginx" {
 
   name = "production-nginx"
   environment = "production"
-  domain = "production.marketplace.team"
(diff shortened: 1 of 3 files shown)


Wynndow:
[Import production DNS records into terraform](https://trello.com/c/xE0dflgq/132-import-production-dns-records-into-terraform)

PR #274, opened 2017-05-12
Open production environment for all IPs

allait:
user_ips are set in common credential variables since they're the
same for all environments apart from production.

We could override the value with the production variable file, but
it making production explicitly set 0.0.0.0/0 as allowed IPs seems
more straightforward and less prone to error if the config file
load order changes for some reason.

### DIFF
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -39,7 +39,7 @@ module "production_nginx" {
 
   admin_user_ips = "${var.admin_user_ips}"
   dev_user_ips = "${var.dev_user_ips}"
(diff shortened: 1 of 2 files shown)


terraform/environments/production/main.tf
```diff
@@ -39,7 +39,7 @@ module "production_nginx" {
 
   admin_user_ips = "${var.admin_user_ips}"
   dev_user_ips = "${var.dev_user_ips}"
-  user_ips = "${var.user_ips}"
+  user_ips = ["0.0.0.0/0"]
```
Wynndow: Is it worth pulling this back to the environment specific vars files at all?

PR #281, opened 2017-05-16
User email domain change to `marketplace.team`

idavidmcdonald:
This will mean our email addresses for our users can be added to mailchimp lists. This is important so we can run scripts that add users to mailchimp lists on our preview and staging environments.

We can no longer use @example.com to do this because mailchimp does not allow emails with this domain to be added as it affects its delivery rate (the same reason you can not subscribe a google
group to a mailchimp list under the group email).

Related trello story where we wish to add user emails to mailchimp lists: https://trello.com/c/XOCeuOYW/254-update-mailchimp-user-list-everyday

### DIFF
--- a/dmaws/rds.py
+++ b/dmaws/rds.py
@@ -239,7 +239,7 @@ class RDSPostgresClient(object):
             bcrypt.gensalt(4)
         ).decode('utf-8')
         self.cursor.execute(
-            "UPDATE users SET name = 'Test user', email_address = id || '-user@example.com', password = '{}'"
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Would be nice if we have a catch-all mailbox set up on `marketplace.team` so that we can see things that we send there, and so we don't get hard failures from e.g. Mandrill and Notify if we send stuff there by mistake/on purpose.

TheDoubleK:
Core team business, that... @allait @Wynndow

PR #288, opened 2017-05-17
Bump production API instance size to 2GB

allait:
Getting a list of framework suppliers seems to run into the memory
limits on production.


TheDoubleK:
:+1:

TheDoubleK:
This is an endpoint used in the admin app, not just for scripts :(

PR #292, opened 2017-05-26
Add script to remove old AMI's and snapshots

Wynndow:
AMIs and their snapshots are not free to keep, so we should clean up the
old ones. When we update the AMIs for our main nginx group and the
elasticsearch cluster we need to remove the old ones.

The script will fetch data about all AMIs that begin with either `nginx` or
`elasticsearch` and are owned by the main account. It sorts them by
creation date and then drops that last AMI (which will be the one we are
currently using). From that data it extracts the ImageId and the
`SnapshotId`. The `ImageId` is used to deregister the image, then then the
`SnapshotId` is used to delete the snapshot.

AMIs must be deregistered before their snapshot can be removed. An
unregisterd AMI will still appear in the list returned by
`describe-images` for some time after being deregistered, which is why
the script can still fetch the `SnapshotId`s after the images have been
deregistered.

Before removing old AMI's it's important to make sure that they are not
currently being used on preview, staging or production. If they are, and
the image is removed, the group will no longer be able to scale.


Wynndow:
[Clean up old AWS environment stuff](https://trello.com/c/4C1bPgX9/126-clean-up-old-aws-environment-stuff)

PR #293, opened 2017-05-26
Add hostedgraphite role

sakisv:
This is to allow hosted graphite pull logs from the AWS account as
described [here](https://www.hostedgraphite.com/docs/integrationguide/ig_aws_cloudwatch.html)


terraform/modules/iam-common/hostegraphite.tf
```diff
@@ -0,0 +1,54 @@
+resource "aws_iam_policy" "hostedgraphite" {
+  name = "Hostedgraphite"
+  policy = <<EOF
+{
+    "Statement": [
+        {
+            "Sid": "PermissionsForMetrics",
+            "Effect": "Allow",
+            "Action": [
```
allait: Let's keep cloudwatch and ec2 permissions, everything else we don't have or wouldn't want in Grafana anyway.

terraform/modules/iam-common/hostegraphite.tf
```diff
@@ -0,0 +1,54 @@
...
+                "kinesis:ListStreams",
+                "redshift:DescribeClusters",
+                "cloudfront:ListDistributions"
+            ],
+            "Resource": [ "*" ]
+        },
+        {
+            "Sid": "PermissionsForTags",
+            "Effect": "Allow",
+            "Action": [
```
allait: Same here, elb permissions are fine, everything else we shouldn't need

sakisv:
@allait done. I am a bit concerned as to whether hostedgraphite will handle the lack of permissions gracefully, but I don't see a reason why it wouldn't.

PR #294, opened 2017-05-30
Bump preview and staging api memory to 2GB

Wynndow:
Downloading the buyers user list from the admin app is still breaking
on preview and staging. Bumping to match production where it works.


TheDoubleK:
Production vars also specify `instance_type: t2.medium` under `api:` - do we also need that here?  Or just the memory should do it?

Wynndow:
Just memory should do it. I'm not 100% sure where those are used, but it's not in the manifests for the PaaS apps which is where the memory size is defined.

TheDoubleK:
Sounds good - I trust you :+1:

PR #298, opened 2017-06-09
Raise log streaming lambda function timeout to 10s

allait:
We've seen some occasional (a few per day) timeouts on the log
streaming lambda functions likely caused by slow Elasticsearch
responses.

Raising the timeout value from 5s to 10s might help reduce the number
of failed runs. We don't want a very high timeout value since if
Elasticsearch at any point starts to respond to requests very slowly
our functions will spend more time waiting which can very quickly
raise the AWS Lambda costs.


idavidmcdonald:
👍

PR #306, opened 2017-06-15
Add briefs frontend to application logs module

Wynndow:
The new briefs frontend needs application logs setting up, the same way
we do for the other apps.

### DIFF
--- a/terraform/modules/log-streaming/subscriptions.tf
+++ b/terraform/modules/log-streaming/subscriptions.tf
@@ -4,20 +4,38 @@ data "aws_region" "current" {
 
 data "aws_caller_identity" "current" {}
 
-resource "aws_lambda_permission" "cloudwatch" {
-  count = "${length(var.log_groups)}"
-  statement_id = "cloudwatch-lambda-${element(var.log_groups, count.index)}"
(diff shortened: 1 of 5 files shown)


terraform/modules/log-streaming/subscriptions.tf
```diff
@@ -4,20 +4,38 @@ data "aws_region" "current" {
...
+  function_name = "${aws_lambda_function.log_stream_lambda.arn}"
+  principal = "logs.${data.aws_region.current.name}.amazonaws.com"
+  source_arn = "arn:aws:logs:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:log-group:${element(var.application_log_groups, count.index)}:*"
+}
+
+resource "aws_cloudwatch_log_subscription_filter" "elasticsearch_subscription_application" {
+  count = "${length(var.application_log_groups)}"
+  name = "elasticsearch-subscription-${element(var.application_log_groups, count.index)}"
+  log_group_name = "${element(var.application_log_groups, count.index)}"
+  filter_pattern = "{ $.url != \"*/_status?ignore-dependencies\" }"
```
allait: I think we might still need `$.url NOT EXISTS` here as not all app logs might have the field

PR #313, opened 2017-06-30
Add simple script to automate dashboard creation

sakisv:
The script loops over the contents of
`scripts/templates/grafana-dashboards/` and sends each of them to
hostedgrapghite to generate the corresponding dashboard.

Using the url scheme suggested in the [docs](https://www.hostedgraphite.com/docs/advanced/grafana-api.html),
i.e. embedding the api key in the url, resulted in 401s.

### DIFF
--- a/scripts/create-hosted-graphite-dashboards.py
+++ b/scripts/create-hosted-graphite-dashboards.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Description:
+    To add
+
+Usage:
+    scripts/create-hosted-graphite-dashboards.py <hosted_graphite_api_key>
+
+Example:
+    scripts/create-hosted-graphite-dashboards.py apikey
+"""
+
+import os
+import sys
+
+import requests
+from docopt import docopt
+
+sys.path.insert(0, '.')  # noqa
+
+
+def generate_dashboards(api_key):
+    endpoint = "https://api.hostedgraphite.com/api/v2/grafana/dashboards/"
+    path = "scripts/templates/grafana-dashboards/"
+
+    for filename in os.listdir(path):
+        with open(path + filename) as fp:
+            resp = requests.put(endpoint, auth=(api_key, ''), data=fp.read())
+            resp.raise_for_status()
+
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__)
+    generate_dashboards(arguments['<hosted_graphite_api_key>'])
--- a/scripts/create-hosted-graphite-dashboards.py
+++ b/scripts/create-hosted-graphite-dashboards.py
@@ -22,7 +22,7 @@ sys.path.insert(0, '.')  # noqa
 
 def generate_dashboards(api_key):
     endpoint = "https://api.hostedgraphite.com/api/v2/grafana/dashboards/"
-    path = "scripts/templates/grafana-dashboards/"
+    path = "../grafana/"
 
     for filename in os.listdir(path):
         with open(path + filename) as fp:
--- a/scripts/templates/grafana-dashboards/dm-apps.json
+++ b/scripts/templates/grafana-dashboards/dm-apps.json
@@ -1,0 +1,747 @@
+{
+  "title": "DM-Apps",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "sharedCrosshair": false,
+  "hideControls": false,
+  "time": {
+    "from": "now-24h",
+    "to": "now"
+  },
+  "timepicker": {
+    "refresh_intervals": [
+      "5s",
+      "10s",
+      "30s",
+      "1m",
+      "5m",
+      "15m",
+      "30m",
+      "1h",
+      "2h",
+      "1d"
+    ],
+    "time_options": [
+      "5m",
+      "15m",
+      "1h",
+      "6h",
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "templating": {
+    "list": [
+      {
+        "allValue": null,
+        "current": {
+          "tags": [],
+          "text": "production",
+          "value": "production"
+        },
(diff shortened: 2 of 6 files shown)


scripts/templates/grafana-dashboards/dm-apps.json
```diff
@@ -0,0 +1,806 @@
...
+        "current": {
+          "selected": true,
+          "text": "api-release",
+          "value": "api-release"
+        },
+        "hide": 0,
+        "includeAll": false,
+        "label": null,
+        "multi": false,
+        "name": "appRelease",
```
allait: Can we remove this now that we've started renaming `-release` metrics?

scripts/templates/grafana-dashboards/dm-apps.json
```diff
@@ -0,0 +1,806 @@
...
+            }
+          ],
+          "span": 12,
+          "stack": true,
+          "steppedLine": false,
+          "targets": [
+            {
+              "hide": false,
+              "refId": "A",
+              "target": "alias(cloudwatch.request_time_buckets.$environment.$app.request_time_bucket_0.samplecount, '0-0.025s')"
```
allait: I think all of these should be `.samplecount:sum`.  Averages aren't very interesting in this case, especially when it comes to 500s and slow requests

scripts/templates/grafana-dashboards/dm-overview.json
```diff
@@ -0,0 +1,510 @@
...
+            {
+              "hide": false,
+              "refId": "J",
+              "target": "alias(cloudwatch.request_time_buckets.$environment.$app.request_time_bucket_9.samplecount, '>10s')"
+            }
+          ],
+          "thresholds": [],
+          "timeFrom": null,
+          "timeShift": null,
+          "title": "Stacked request times $environment $app",
```
allait: The bars shows request counts, so maybe "Request counts grouped by time"?

scripts/templates/grafana-dashboards/dm-overview.json
```diff
@@ -0,0 +1,509 @@
...
+          {
+            "selected": true,
+            "text": "production",
+            "value": "production"
+          }
+        ],
+        "query": "preview,staging,production",
+        "type": "custom"
+      },
+      {
```
allait: Why does the main dashboard need the app selector? It should be one dashboard per environment. I don't remember if we've talked about it, but I think it should show nginx requests and slow requests, api requests and slow requests, nginx 500s and combined memory (and maybe cpu) metrics across all apps.

scripts/templates/grafana-dashboards/dm-overview.json
```diff
@@ -0,0 +1,509 @@
+{
+  "title": "DM Overview",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
+  "sharedCrosshair": false,
+  "hideControls": false,
+  "time": {
+    "from": "now-6h",
```
allait: Can we make this 12h or 24h? 6h is not very useful in the morning.

scripts/templates/grafana-dashboards/dm-apps.json
idavidmcdonald: So using `sum` or `avg` on these is for an individual metric at a timestamp rather than any aggregation of a metric over a period of time.

For example if we have
```
requests_bucket_1 50 1500031162
requests_bucket_1 100 1500031162
```
then using `sum`  would give us
```requests_bucket_1 150 1500031162```
or using `avg` would give us
```requests_bucket_1 75 1500031162```

We did had a problem that we were sending different values for the same metric for the same timestamp (first an incomplete value and then the full value) so would make a difference if we chose sum or avg. However, we have fixed this so there is no duplicated metric data being sent and therefore it should not matter if we choose sum or avg. 

Therefore it feels like we are fine to keep avg. If we really wanted to cover the case were we would have double metrics being sent per timestamp then we could use the `max` filter.

File: scripts/templates/grafana-dashboards/dm-apps.json [hunk 6]
```diff
@@ -0,0 +1,747 @@
+{
+  "title": "DM-Apps",
+  "tags": [],
+  "style": "dark",
+  "timezone": "browser",
+  "editable": true,
```
allait: If we set this to `false` it removes the Edit buttons and prevents the dashboards from being modified in the interface, so I think we should do that.

scripts/templates/grafana-dashboards/dm-apps.json
```diff
@@ -0,0 +1,747 @@
+{
+  "title": "DM-Apps",
```
allait: Very minor, but let's drop the dash like in "DM Overview"

scripts/create-hosted-graphite-dashboards.py
```diff
@@ -0,0 +1,35 @@
...
+
+import requests
+from docopt import docopt
+
+sys.path.insert(0, '.')  # noqa
+
+
+def generate_dashboards(api_key):
+    endpoint = "https://api.hostedgraphite.com/api/v2/grafana/dashboards/"
+    path = "scripts/templates/grafana-dashboards/"
```
allait: I'd maybe think about moving the templates to a top-level directory alongside kibana (like `graphite` or `grafana`) and adding it to the README as it took me a bit to find it in the checkout.

scripts/templates/grafana-dashboards/dm-apps.json
sakisv: Ok, just to point out, this only changes whether or not you can change the rows. You will still be able to move graphs around or change what is displayed in them, just you'll not be able to add and change the rows.

allait: Does it? For me it removes the "Edit" buttons from the graphs. I can still drag things around on the dashboard but when I leave it doesn't offer to save any changes

sakisv: Weird.. Well, I guess we'll find out soon 😄

PR #315, opened 2017-07-24
Create zipped and encrypted DB backups and upload to S3

Wynndow:
This adds a new app, which when deployed will connect to the database, use `pg_dump` to grab a dump of it. Before writing to disk, the dump is zipped and then encrypted. The file is then uploaded to S3 for storage.

The original plan was to take the dump for a brand new postgres service created from the latest snapshot with PaaS take. Unfortunately there are some issues with creating these new postgres services on PaaS's side. They're working on a fix. When that happens, the app can easily be configured to use a new service. The Jenkins job which creates and deploys the app will also need updating to use the `make` rules being commited here.

The encryption uses GPG2. The public key is being committed here. The private key has been added to the credentials repo in a separate pull request. The key is RSA 4096 bits. As the docker image is only being used for encryption, it doesn't need a key of its own and can just use the imported public key.

The dumps are being uploaded to s3 via a signed URL. This url needs to be created on the Jenkins box, as the role is assumes has the correct permissions for uploading files. These details are then injected into the new app so it can use them.


db-backup/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.6.1-slim
+
+COPY /create-db-dump.sh /tmp/create-db-dump.sh
```
allait: `/tmp` feels like a weird location for these, maybe `/` or `/app`

db-backup/create-db-dump.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
+
+pg_dump "${DB_URI}" --no-acl --no-owner | gzip | \
+  gpg2 --trust-model always -a -r 'Digital Marketplace DB backups' --out /tmp/dump.sql.gz.gpg --encrypt
```
allait: Maybe make `'Digital Marketplace DB backups'` an environment variable value

db-backup/create-db-dump.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
+
+pg_dump "${DB_URI}" --no-acl --no-owner | gzip | \
```
allait: I'd be interested to see if the custom format makes a big difference in terms of memory usage, size or speed of backup / restore, but that's something we can tweak later

requirements.txt
```diff
@@ -1,7 +1,7 @@
 click==6.6
 Jinja2==2.8
 PyYAML==3.11
-
 six==1.10.0
-
-awscli==1.11.44
+awscli==1.11.4
```
allait: typo, should still be `.44`

terraform/accounts/main/s3_buckets.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_s3_bucket" "database_backups_s3_bucket" {
+  bucket = "digitalmarketplace-database-backups"
+  acl    = "private"
+
+  versioning {
+    enabled = true
```
allait: If terraform complains in the future we'll need to add `mfa_delete = true` to this once we've switched it on.

scripts/check-db-dump-is-decryptable.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+LATEST_DUMP=$(aws s3 ls s3://digitalmarketplace-database-backups | sort | tail -n 1 | awk '{print $4}')
```
allait: We generate the file name for `S3_POST_URL_DATA`, is there a way for us to store it and use that? It would allow us to use the script to check any backup as well as making sure we're checking the thing we've just created and not some other file that happens to have a later date for some reason.

PR #317, opened 2017-08-04
Clean up production database dumps and migrate to other stages

Wynndow:
See also this PR on Jenkins: [https://github.com/alphagov/digitalmarketplace-jenkins/pull/18](https://github.com/alphagov/digitalmarketplace-jenkins/pull/18)

Adds scripts and make rules for use by Jenkins to clean up the nightly production database dumps of sensitive information, and then migrate to staging or preview, as well as upload to a shared directory on google drive.

To clean the dump a new service is created in a separate space, db-cleanup, as well as a skeleton app to allow an ssh tunnel through to the database service. The latest database dump is then imported and a series of sql commands run against it to remove sensitive info.

That cleaned db is then downloaded back to Jenkins and then migrated to the desired environments database, as well as uploading to google drive.


scripts/migrate-cleaned-db-to-target-stage.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+set -euxo pipefail
```
katstevens: Should we set `-x` on all the pipefails?

Makefile
```diff
@@ -106,7 +106,7 @@ create-db-snapshot-service: ## Create a db service from the latest db snapshot
 
 .PHONY: check-db-snapshot-service
 check-db-snapshot-service: ## Get the status for the db snapshot service
-	@cf service digitalmarketplace_api_db_snapshot | grep -i 'status: ' | sed 's/^.*: //' | awk '{print toupper($0)}'
+	@cf service digitalmarketplace_api_db_snapshot | grep -i 'status: ' | sed 's/^.*: //' | tr '[:lower:]' '[:upper:]'
```
katstevens: Nice - I haven't seen `tr` used before!

scripts/clean-db-dump.sql
```diff
@@ -0,0 +1,102 @@
...
+
+\echo 'Delete supplier frameworks for open frameworks'
+DELETE
+FROM supplier_frameworks
+WHERE framework_id IN
+    (SELECT id
+     FROM frameworks
+     WHERE status='open');
+
+-- Remove data related to open DOS procurements
```
katstevens: Question: are we using `DELETE CASCADE` by default on our DB?

Makefile
```diff
@@ -133,6 +133,37 @@ cleanup-db-backup: ## Remove snapshot service and app
...
+.PHONY: migrate-cleaned-db-dump
+migrate-cleaned-db-dump: ## Migrate the cleaned db dump to a target stage and sync with google drive.
+	@[ $$(cf target | grep -i 'space' | cut -d':' -f2) = "db-cleanup" ] || (echo "Error: This can only be run in the db-cleanup space" && exit 1)
+	./scripts/migrate-cleaned-db-to-target-stage.sh
+
+.PHONE: cleanup-db-cleanup
+cleanup-db-cleanup: ## Delete app and service created in the db cleanup procedure
+	@[ $$(cf target | grep -i 'space' | cut -d':' -f2) = "db-cleanup" ] || (echo "Error: This can only be run in the db-cleanup space" && exit 1)
+	cf delete -f db-cleanup
+	cf delete-service -f digitalmarketplace_db_cleanup
```
katstevens: What happens when the production data is a migration (or two) behind preview/staging? Is this handled in a later script or make step?

scripts/create-db-cleanup-service.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+cf target -s preview
+DB_PLAN=$(cf service digitalmarketplace_api_db | grep -i 'plan: ' | cut -d ' ' -f2)
+cf target -s db-cleanup
+[ $(cf target | grep -i 'space' | cut -d':' -f2) = "db-cleanup" ] || (echo "Error: This can only be run in the db-cleanup space" && exit 1)
+cf create-service postgres ${DB_PLAN} digitalmarketplace_db_cleanup
```
Wynndow: Yeah, I think the `sleep 10`s thing was me misunderstanding cloud foundry and wasn't actually needed. When switching spaces it turns out it does happen in a subshell which only returns when it's done, so it's all good.

scripts/migrate-cleaned-db-to-target-stage.sh
Wynndow: Whoops, I meant to take this out. I was using this for debugging!

Makefile
Wynndow: This was the first time I heard of/used it. Stack overflow is a wonderful thing.

Makefile
Wynndow: That's a good question.

I guess that when the clean prod data is migrated to the stage, it could break? In that scenario we would could re-release the latest api release up to preview/staging which will run the latests migrations against the new data?

scripts/clean-db-dump.sql
Wynndow: I don't know to be honest, I'm not super familiar with it. The SQL here is pretty much a copy/paste job from our old cleanup script.

Would it be beneficial?

Makefile
katstevens: Thinking out loud: let's say preview contains some new api code with a migration for e.g. an altered column. If we feed in the cleaned prod data dump before we build the app and run those migrations - e.g. for a dev getting set up locally - it would likely be fine. But if we're arbitrarily updating the preview DB on a Sunday night, we should first check whether the migration version is in sync or not. If preview has a later alembic version stamp than prod, we can roll back to the prod version, apply the dump, then roll forward again to the preview head version. What do you think?

scripts/clean-db-dump.sql
katstevens: I think using `CASCADE` is slightly more efficient for deleting child objects/avoiding orphans - so e.g. deleting draft `Brief` objects would also delete their child `BriefUser` objects without needing a separate statement. However it's obviously less explicit what's being deleted and what's not. If the script above has been serving us well in the past then I'm happy for it to continue 😄

Makefile
Wynndow: Yeah, that makes sense. Do we need to roll back? If preview has a later version stamp we could just run the migrations immediately after applying the dump?

scripts/clean-db-dump.sql
```diff
@@ -0,0 +1,102 @@
...
+WHERE brief_id IN
+    (SELECT id
+     FROM briefs
+     WHERE published_at IS NULL);
+
+\echo '  > Delete draft briefs'
+DELETE
+FROM briefs
+WHERE published_at IS NULL;
+
```
benvand: This could be a little clearer if it was defined as some plpgsql functions/ variables but I gather this is mostly a copy paste from another script right? In that case probably out of scope to rewrite all this?

scripts/clean-db-dump.sql
```diff
@@ -0,0 +1,102 @@
...
+\echo 'Delete framework agreements'
+DELETE
+FROM framework_agreements;
+
+-- Overwrite declarations with the smallest possible valid entry
+-- Removes all personal data while keeping our app working as expected
+\echo 'Blank out declarations'
+UPDATE supplier_frameworks
+SET declaration = (CASE
+                       WHEN (declaration->'status') IS NULL
```
benvand: Might be worth dropping the case declaration onto a new line for readability
```
SET declaration = (
    CASE
         WHEN (declaration->'status') IS NULL OR (declaration->'nameOfOrganisation') IS NULL 
         THEN '{}'
         ELSE '{
             "status": "' || (declaration->>'status') || '",
             "nameOfOrganisation": "' || replace((declaration->>'nameOfOrganisation'), '"', '') || '",
             "primaryContactEmail": "supplier-user@example.com",
             "organisationSize": "' || (ARRAY['micro', 'small', 'medium', 'large'])[MOD(supplier_id, 4)+1] || '" 
```

scripts/generate-bcrypt-hashed-password.py
```diff
@@ -0,0 +1,29 @@
...
+
+Usage:
+scripts/generate-bcrpyt-hashed-password.py <password> <cost-factor>
+"""
+
+import bcrypt
+from docopt import docopt
+
+
+def hash_password(password, cost_factor):
```
benvand: We can fit all that on one line and not declare it as a variable. Line limit should be at 120.
`return bcrypt.hashpw(bytes(password), bcrypt.gensalt(cost_factor)).decode('utf-8')`

Makefile
katstevens: Oops, yes, you're right - `psql` won't care about the 'new' structure yet. This also has the advantage of nicely simulating what will happen when we apply any complicated migrations on prod 😄

Makefile
```diff
@@ -133,6 +133,37 @@ cleanup-db-backup: ## Remove snapshot service and app
...
+.PHONY: deploy-db-cleanup-app
+deploy-db-cleanup-app: ## Deploys the db cleanup app
+	@[ $$(cf target | grep -i 'space' | cut -d':' -f2) = "db-cleanup" ] || (echo "Error: This can only be run in the db-cleanup space" && exit 1)
+	cf push db-cleanup -o digitalmarketplace/db-cleanup --no-route --health-check-type none -i 1 -m 128M -c 'sleep 2h'
+	cf bind-service db-cleanup digitalmarketplace_db_cleanup
+	cf restage db-cleanup
+
+.PHONY: import-and-clean-db-dump
+import-and-clean-db-dump: virtualenv ## Connects to the db-cleanup service, imports the latest dump and cleans it.
+	@[ $$(cf target | grep -i 'space' | cut -d':' -f2) = "db-cleanup" ] || (echo "Error: This can only be run in the db-cleanup space" && exit 1)
```
Wynndow: We actually have command that does this already in the Makefile, I'll update to use.

scripts/clean-db-dump.sql
Wynndow: Yeah, it's the same SQL as we used before. How much work would it be?

scripts/clean-db-dump.sql
Wynndow: Cool. I was trusting by beautifier too much I think!

scripts/generate-bcrypt-hashed-password.py
Wynndow: Nice.

PR #319, opened 2017-08-14
Sanitise rather than delete framework agreements

TheDoubleK:
We can sanitise the personal details (signer name and role, and the S3
paths to actual signed/countersigned documents) but preserve the fact that
the agreement has been returned.

This means that when logging in as a supplier in Preview or Staging
environments the framework dashboard will be accessible without having
to go through the whole "return your framework agreement" flow.

This will be useful for developers working on the framework dashboard
who do not want to repeatedly return fake framework agreements for
suppliers in order to test their new features.

### DIFF
--- a/scripts/clean-db-dump.sql
+++ b/scripts/clean-db-dump.sql
@@ -75,10 +75,21 @@ WHERE supplier_id IN
 DELETE
 FROM audit_events;
 
--- Remove framework agreements because they contain personal data and we don't rely on them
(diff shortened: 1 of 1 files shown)


TheDoubleK:
I'm about 95% happy this will work alright.  Not sure if the non-existent S3 paths might cause an issue on some pages, but reckon it's worth a suck-it-and-see.

PR #320, opened 2017-08-16
Add brief responses frontend

Wynndow:
### Add brief-responses-frontend to list of apps in terraform
Adding the new app to the list in the application logs module will
create the log groups as well as the metrics.

### Add brief-responses-frontend to vars/common.yml
This sets the domain and the path for the new app when deployed to PaaS.


Wynndow:
@benvand Yeah, that's a good shout. Thanks very much!

PR #322, opened 2017-08-17
Add Kat and Louise users to the terraform module

allait:
User accounts were created manually, so I had to import them into
the Terraform state.


terraform/modules/iam-users/users/main.tf
```diff
@@ -26,6 +26,10 @@ resource "aws_iam_user" "george_lund" {
   name = "georgelund"
 }
 
+resource "aws_iam_user" "kat_stevens" {
+  name = "katstevens"
```
katstevens: 🏆

PR #324, opened 2017-08-21
Add routing for direct-award on buyer-frontend

samuelhwilliams:
## Summary
Adds routing so that buyer-frontend will receive anything that goes to `/buyers/direct-award` and also be the fallback for un-routed urls at `/`. Tweaks templating to allow multiple paths to be supplied. The first path specified in `common.yml` against the app will be the one used for the healthcheck. Not sure if this is preferrable to explicitly adding another key called e.g. `healthcheck_endpoint`, but will leave that open to review.

Also not sure what you do so far as testing, but I didn't see anything pre-existing to edit or add to. Did I just miss it, or is it in another repo, or..?

## Ticket
https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/paas/_base.j2
+++ b/paas/_base.j2
@@ -3,12 +3,14 @@
 applications:
   - name: {{ app|replace('_', '-') }}-release
     routes:
-      - route: {{ subdomain }}-{{ environment }}.cloudapps.digital{{ path|default('') }}
+{% for path in (paths|default([''])) %}
+      - route: {{ subdomain }}-{{ environment }}.cloudapps.digital{{ path }}
+{% endfor %}
     instances: {{ instances|default(1) }}
     memory: {{ memory|default('512M') }}
     disk_quota: {{ disk_quota|default('1024M') }}
     health-check-type: http
-    health-check-http-endpoint: {{ path|default('') }}/_status?ignore-dependencies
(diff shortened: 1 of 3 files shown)


vars/common.yml
```diff
@@ -9,18 +9,24 @@ search-api:
...
+  paths:
+    - /suppliers
 
 briefs-frontend:
   subdomain: dm
-  path: /buyers
+  paths:
+    - /buyers
 
 brief-responses-frontend:
```
Wynndow: I think the path for the brief-responses app needs moving to the array pattern like the others.

paas/_base.j2
```diff
@@ -3,12 +3,14 @@
 applications:
   - name: {{ app|replace('_', '-') }}-release
     routes:
+{% for path in (paths|default([''])) %}
       - route: {{ subdomain }}-{{ environment }}.cloudapps.digital{{ path|default('') }}
```
Wynndow: Can you simplify to just `{{ path }}`?

samuelhwilliams:
@allait better?

vars/common.yml
samuelhwilliams: Good spot. Thanks.

PR #328, opened 2017-09-08
Fix "string argument without an encoding" python3 error in bcrypt script

allait:
generate-bcrypt-hashed-password raises an error on python3 since `bytes` is called without an encoding argument. Replacing it with `.encode` should fix the problem.

### DIFF
--- a/scripts/generate-bcrypt-hashed-password.py
+++ b/scripts/generate-bcrypt-hashed-password.py
@@ -16,7 +16,7 @@ from docopt import docopt
 
 
 def hash_password(password, cost_factor):
-    return bcrypt.hashpw(bytes(password), bcrypt.gensalt(cost_factor)).decode('utf-8')
+    return bcrypt.hashpw(password.encode('utf-8'), bcrypt.gensalt(cost_factor)).decode('utf-8')
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


scripts/generate-bcrypt-hashed-password.py
```diff
@@ -16,7 +16,7 @@
 
 
 def hash_password(password, cost_factor):
-    return bcrypt.hashpw(bytes(password), bcrypt.gensalt(cost_factor)).decode('utf-8')
+    return bcrypt.hashpw(password.encode('utf-8'), bcrypt.gensalt(cost_factor)).decode('utf-8')
```
galund: Noticed the 'decode' call at the end of this: is there any guarantee that bcrypt returns a string that can be interpreted as UTF-8? Isn't it just an arbitrary byte-string at this point? Presumably we're passing straight on the a URI-encoding function, in which case it might be better if this function did just return bytes to be fed in there, rather than a unicode string?

allait: bcrypt hashes contain a base64 string and `$` signs for version/cost factor, so they're safe to decode. We're printing them out right away, so I don't think we need to decode, but it can be useful if the function will end up being used anywhere else.

PR #329, opened 2017-09-08
Fix signed agreements cleanup

allait:
##  Interrupt and exit the clean DB dump SQL script on error

Currently, if any of the SQL commands in the clean DB dump script
fail to apply psql will ignore the error and continue, existing with
a 0 status in the end. This could result in DB dumps that still
contain personal data if eg there's a syntax error in one of the
clean-up commands.

ON_ERROR_STOP will terminate psql and exit with a non-zero status if
any of the SQL commands return with an error.


TheDoubleK:
Damn, my bad there I think. 

:+1:

PR #330, opened 2017-09-11
Create closed brief responses

benvand:


scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,47 @@ WHERE declaration IS NOT NULL
...
+)
+SELECT
+    '{"essentialRequirements": [{"evidence": "They have the correct number of hooves."}, {"evidence": "So shiny..."}, {"evidence": "All at least 50"}], "niceToHaveRequirements": [{"yesNo": false}, {"yesNo": true, "evidence": "No jump is too high."}, {"yesNo": true, "evidence": "NEIGH"}], "availability": "09/09/17", "essentialRequirementsMet": true, "respondToEmailAddress": "example-email@gov.uk"}'::json,
+    eligible_brief_supplier_pairings.brief_id,
+    eligible_brief_supplier_pairings.supplier_id,
+    CASE
+      WHEN MOD(id, 5) = 0 THEN NULL
+      ELSE now()
+    END,
+    now()
```
katstevens: This might be my rusty SQL knowledge showing, but does this always put `now()` into `submitted_at` and the case statement into `created_at`? Shouldn't it be the other way around?

benvand:
@katstevens good catch

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,36 @@ WHERE declaration IS NOT NULL
   AND declaration::varchar != 'null'
   AND declaration::varchar != '{}';
 
+-- Create some fake draft and closed briefs using eligible suppliers
```
idavidmcdonald: This line is no longer up to date

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,36 @@ WHERE declaration IS NOT NULL
...
+      created_at,
+      submitted_at
+)
+SELECT
+    '{"essentialRequirements": [{"evidence": "They have the correct number of hooves."}, {"evidence": "So shiny..."}, {"evidence": "All at least 50"}], "niceToHaveRequirements": [{"yesNo": false}, {"yesNo": true, "evidence": "No jump is too high."}, {"yesNo": true, "evidence": "NEIGH"}], "availability": "09/09/17", "essentialRequirementsMet": true, "respondToEmailAddress": "example-email@gov.uk"}'::json,
+    eligible_brief_supplier_pairings.brief_id,
+    eligible_brief_supplier_pairings.supplier_id,
+    now(),
+    CASE
+      WHEN MOD(id, 5) = 0 THEN NULL
```
idavidmcdonald: Nice use of MOD

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,36 @@ WHERE declaration IS NOT NULL
...
+-- Create some fake draft and closed briefs using eligible suppliers
+INSERT INTO brief_responses (
+      data,
+      brief_id,
+      supplier_id,
+      created_at,
+      submitted_at
+)
+SELECT
+    '{"essentialRequirements": [{"evidence": "They have the correct number of hooves."}, {"evidence": "So shiny..."}, {"evidence": "All at least 50"}], "niceToHaveRequirements": [{"yesNo": false}, {"yesNo": true, "evidence": "No jump is too high."}, {"yesNo": true, "evidence": "NEIGH"}], "availability": "09/09/17", "essentialRequirementsMet": true, "respondToEmailAddress": "example-email@gov.uk"}'::json,
```
idavidmcdonald: The number of essential requirements and n2h requirements may not be equal to the number the brief is looking for. Can you think of any implications of this? 

For example, I think... if someone logged in as the buyer of a random brief and tried to download the responses, it might fail. 

Trying to figure out if this is acceptable or not (it might be a fair bit harder to create the correct number of pieces of evidence)

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,36 @@ WHERE declaration IS NOT NULL
...
+    END
+FROM (
+    SELECT DISTINCT ON (briefs.id)
+      briefs.id AS brief_id,
+      supplier_id as supplier_id
+    FROM supplier_frameworks
+      LEFT JOIN frameworks ON supplier_frameworks.framework_id = frameworks.id
+      LEFT JOIN briefs ON briefs.framework_id = frameworks.id
+    WHERE declaration->>'status' = 'complete'
+      AND declaration->>'organisationSize' != ''
```
idavidmcdonald: I'm not too up to date with services, but is there ever a reason why the declaration would be complete but they don't have an organisation size?

TheDoubleK: Copied declarations should all have a value here - see https://github.com/alphagov/digitalmarketplace-aws/blob/master/scripts/clean-db-dump.sql#L108

benvand: Live also all have it too as it's been a required value on all declarations.

scripts/clean-db-dump.sql
benvand: I think that provisioning for this would take a lot of work.
```
digitalmarketplace=# select len, count(len) from (select json_array_length((data->>'essentialRequirements')::json) as len from briefs) as len_table group by len;
 len | count 
-----+-------
     |     0
   8 |    82
  16 |    16
  15 |    22
   4 |  3172
  20 |    13
   1 |    40
  13 |    41
   5 |   121
  11 |    49
   3 |   135
  14 |    26
  17 |    18
  19 |    20
  12 |    39
  10 |    96
  18 |    21
   9 |   133
   6 |   117
   2 |   126
   7 |   112
```

The majority of briefs have 4 essential requirements. I could try and find out if there's some sql to make them match but I can't think of any off the top of my head.

benvand: Ah I'd need to add another essential requirement to the test data for a start.

scripts/clean-db-dump.sql
benvand: Found a way. In terms of creating test data I think it'll do. Creates an `essentialRequirements` with the same number of 'evidences' as there are essential requirements.

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,38 @@ WHERE declaration IS NOT NULL
...
+-- Ensuring essential requirements evidence length matches that of the brief essential requirements
+INSERT INTO brief_responses (git push 0-family()
+      data,
+      brief_id,
+      supplier_id,
+      created_at,
+      submitted_at
+)
+SELECT
+    ('{"essentialRequirements": [' || rtrim(repeat('{"evidence": "Some super good evidence."}, ', essential_requirements_len), ', ') || '], "niceToHaveRequirements": [{"yesNo": false}, {"yesNo": true, "evidence": "No jump is too high."}, {"yesNo": true, "evidence": "NEIGH"}], "availability": "09/09/17", "essentialRequirementsMet": true, "respondToEmailAddress": "example-email@gov.uk"}')::json,
```
idavidmcdonald: Nice on the essential requirements length. Can we also do it for the nice to have requirements as they can be off various length?

idavidmcdonald: Also is there anyway to split this line over multiple lines to improve the readability whilst still writing valid SQL?

benvand: Done 👍

scripts/clean-db-dump.sql
benvand: Fixed 👍

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,51 @@ WHERE declaration IS NOT NULL
   AND declaration::varchar != 'null'
   AND declaration::varchar != '{}';
 
+-- Create some fake draft and closed brief responses on closed briefs using eligible suppliers
```
idavidmcdonald: Technically the brief response is 'submitted' rather than 'closed'

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,51 @@ WHERE declaration IS NOT NULL
...
+-- Ensuring essential requirements evidence length matches that of the brief essential requirements
```
idavidmcdonald: Could also say the same about nice to have length

scripts/clean-db-dump.sql
```diff
@@ -112,6 +112,51 @@ WHERE declaration IS NOT NULL
...
+)
+SELECT
+    -- Build a string that we convert to a json object for the 'data' column with varying numbers of evidence
+    (
+      '{' ||
+        '"essentialRequirements": [' ||
+          rtrim(repeat('{"evidence": "Some essential evidence."}, ', essential_requirements_len), ', ') ||
+        '], ' ||
+        '"niceToHaveRequirements": [' ||
+          rtrim(repeat('{"yesNo": "true", "evidence": "Some nice to have evidence."}, ', MOD(eligible_brief_supplier_pairings.supplier_id, 6)), ', ') ||
```
idavidmcdonald: So as we are doing MOD 6 on the supplier ID and MOD 5 on the nice to haves am I right in thinking we will have the following cases:
A response which is not submitted which has `niceToHaveRequirements`: []
A response which has been submitted which has `niceToHaveRequirements`: []
A response which is not submitted which has `niceToHaveRequirements`: [correct number of nice to haves]
A response which has been submitted which has `niceToHaveRequirements`: [correct number of nice to haves]

I don't think the first two responses are realistic data that we would see. If they had started an application but not yet answered the nice to have requirements question then we would see no key for nice to have requirements. If they had submitted the application, then they would have answered nice to have requirements and would have data in the list (even if it just said that "yesNo": false). 

You could go to the effort of trying to change the first two cases so you the key only exists with the correct number of nice to haves.
You could also simplify it so that they always have the key full of nice to haves.

Do correct me if any of my interpreting of your SQL is wrong. Also please do double check any of my assumptions if something you decide to change relies on them as this is all of the top of my head.

idavidmcdonald: Should it be `"yesNo": true` rather than `"yesNo": "true"`?

benvand:
🤖

PR #334, opened 2017-09-25
Add notify api key to briefs and supplier frontends

Wynndow:
The briefs and supplier frontends are going to send user create/invite
emails with notify, hence the need for the notify api key.

Briefs frontend will only use notify so we can drop the mandrill key.
Supplier frontend will still use both.

### DIFF
--- a/paas/briefs-frontend.j2
+++ b/paas/briefs-frontend.j2
@@ -8,7 +8,7 @@
       DM_DATA_API_AUTH_TOKEN: {{ api.auth_tokens[0] }}
       DM_DATA_API_URL: https://dm-api-{{ environment }}.cloudapps.digital
 
-      DM_MANDRILL_API_KEY: {{ shared_tokens.mandrill_key }}
+      DM_NOTIFY_API_KEY: {{ notify_api_key }}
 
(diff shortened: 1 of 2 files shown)


paas/briefs-frontend.j2
```diff
@@ -8,7 +8,7 @@
       DM_DATA_API_AUTH_TOKEN: {{ api.auth_tokens[0] }}
       DM_DATA_API_URL: https://dm-api-{{ environment }}.cloudapps.digital
 
-      DM_MANDRILL_API_KEY: {{ shared_tokens.mandrill_key }}
+      DM_NOTIFY_API_KEY: {{ notify_api_key }}
```
TheDoubleK: I'm not decoding credentials to check this, but what's the logic behind moving the token out of the "shared_tokens" section?  Feels like that's exactly what it is.

TheDoubleK:
If you're sure that's the right key then :+1:

Wynndow:
@TheDoubleK It's definitely the right location. Why it's not in the shared keys I don't know. I'm going to merge this to get the tests passing, but can come back to look at it.

PR #337, opened 2017-09-27
Flake8

benvand:
[flake8-report.txt](https://github.com/alphagov/digitalmarketplace-aws/files/1337227/flake8-report.txt)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,11 @@
+[flake8]
+# Rule definitions: http://flake8.pycqa.org/en/latest/user/error-codes.html
+# D203: 1 blank line required before class docstring
+# F401: 'identifier' imported but unused
+# W503: line break before binary operator
+exclude = venv*,__pycache__,node_modules,bower_components
+ignore = D203,W503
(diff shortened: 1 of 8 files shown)


scripts/clean-db-dump.sql
```diff
@@ -144,27 +144,22 @@ SELECT
         '"respondToEmailAddress": "example-email@gov.uk"' ||
       '}'
     )::JSON,
-    eligible_brief_supplier_pairings.brief_id,
+        eligible_brief_supplier_pairings.supplier_id,
```
samuelhwilliams: This doesn't seem like a flake8 change? It looks over-indented, it changes a key, and it duplicates the key below?

samuelhwilliams: Are any of the changes in this file intentional?

benvand: Damn, they snuck in there, good spot.

benvand: Done 👍

PR #341, opened 2017-10-12
Downscale user-frontend and admin-frontend apps

samuelhwilliams:
## Summary
Reviewing monitoring for the last week, our user-frontend and admin-frontend barely hit 2% CPU. Let's scale down the default number of instances we create for these apps.


MuriloDalRi:
What's the default number of instances if you don't specify one?

PR #344, opened 2017-10-20
Preview paas router manifest

allait:
### Remove shell commands from the generate-manifest output

By default, make prints the command being executed, so when the output of generate-manifest is being redirected to a file the resulting manifest contains the command names before the YAML front matter.

Adding `@` stops make from printing the commands, but keeps the command output.

### Explicitly list full routes in PaaS vars

Changes the base manifest template to include application routes as-is (only formatting them to add current environment) instead of constructing the route from domain / subdomain and paths.

This allows us to specify multiple subdomains / domains per app, and specify manifests for apps that don't need a route.

To avoid duplicating the application paths between environments we format routes with the current environment, so they can be specified once in the common vars file.

### Update unmap-route script to support custom domains

PaaS now allows us to add custom domains and map them to existing applications, so route unmapping script should look up the domain for each route instead of always using cloudapps.digital.

### Add PaaS manifest template for the nginx router app

Adds a manifest template and preview variables that would map the subdomains and the CloudFront instance to the nginx router app deployed to PaaS.

This allows us to use the application release pipeline to deploy nginx changes.

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -1,4 +1,15 @@
 ---
+maintenance_mode: live
+
+router:
+  routes:
+    - www.{env}.marketplace.team
+    - api.{env}.marketplace.team
+    - search-api.{env}.marketplace.team
+    - assets.{env}.marketplace.team
+  services:
+    - router_cdn
+
 api:
   routes:
     - dm-api-{env}.cloudapps.digital
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -1,46 +1,41 @@
 ---
 api:
-  subdomain: dm-api
+  routes:
+    - dm-api-{env}.cloudapps.digital
   services:
     - digitalmarketplace_api_db
 
 search-api:
-  subdomain: dm-search-api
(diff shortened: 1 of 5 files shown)


vars/common.yml
```diff
@@ -1,46 +1,52 @@
 ---
+maintenance_mode: live
```
idavidmcdonald: Could you explain this line for me please? It reads like maintenance mode would be on by default and I couldn't find anything else in our aws repo in terms of setting this variable so it seems new

allait: That's the existing nginx `mode` flag from Terraform. When this is set to `maintenance_mode: maintenance` nginx will be deployed with a maintenance page.

The environment variable is just called `DM_MODE`, but I felt like it's worth expanding the name a bit here, even though I agree it's not ideal.

idavidmcdonald: Cheers. Yeah it's not the key that is the issue so much, more the value that can be a little confusing.

PR #348, opened 2017-10-24
Replace nginx with cdn route

allait:
### Add router module to set up DNS record for PaaS cdn-route

Sets up TXT records for ACME domain validation check.
When cname_domain is set to a non-empty value the module also creates the CNAME records to point subdomains to CloudFront.

### Replace environment _nginx resource with _router one

Removes nginx ASG, ELB and error log group. Keeps the Route53 zone, but changes the records from ELB aliases to CloudFront CNAMEs.

JSON log group and Route53 zone have been manually moved using `terraform state mv` from nginx module to router, so that they are not recreated.

### Remove unused ansible playbooks for nginx AMI creation

We're now using https://github.com/alphagov/digitalmarketplace-router to deploy the nginx app.

We're no longer building AMIs for nginx, so we don't need the Ansible playbooks anymore.


packer/jenkins.json
```diff
@@ -1,21 +0,0 @@
...
-  "variables": {
-      "aws_region": "eu-west-1"
-  },
-  "builders": [{
-    "type": "amazon-ebs",
-    "region": "{{ user `aws_region` }}",
-    "source_ami": "ami-d9ba9eae",
-    "instance_type": "t2.micro",
-    "ssh_username": "ubuntu",
-    "ami_name": "jenkins-{{isotime | clean_ami_name}}",
```
idavidmcdonald: We are definitely good to delete this one? I don't entirely understand the packer stuff but obviously it mentions jenkins so that's why I ask

allait: Hmm. Yeah, this got caught in the wide `git rm` net, but maybe it's worth dropping it as well. Essentially the only thing it does is create an AMI with a larger root volume (so instead of the default 8Gb you get 20Gb). Otherwise it's the same as stock ubuntu image.

It was worth keeping the template around, but in practice we've never had to recreate the AMI and would probably restore Jenkins from the EBS snapshot in the future. So as long as that image is not deleted, I'd suggest we drop this template just to remove the overhead of having to know about packer at all.

PR #349, opened 2017-10-25
Do not include `routes` key if no routes are to be added to app

idavidmcdonald:


paas/_base.j2
```diff
@@ -2,10 +2,12 @@
 
 applications:
   - name: {{ app|replace('_', '-') }}-release
+{% if routes|default([]) %}
```
samuelhwilliams: Do we need the default here? I don't think it adds anything when we're doing an `if` check; only needed when iterating with `for`, I'd have thought.

idavidmcdonald: Because the Jinja environment is set up in `StrictUndefined` mode in the generate paas manifest script (http://jinja.pocoo.org/docs/2.9/api/#jinja2.StrictUndefined), the only operation allowed on an undefined object is to check if it is defined. Therefore, if `routes` is not defined and you try `if routes` then we will throw an exception. Therefore we need to use the `default` filter (https://github.com/pallets/jinja/blob/master/jinja2/filters.py#L355) I believe.

samuelhwilliams: Maybe what we want to actually check then is that it's defined? i.e `if routes is defined`?

benvand:
Looks good as a bug fix, should fix the https://ci.marketplace.team/job/database-backup/160/consoleFull job
Is there any reason to not merge the 2 if statements?:

```
{% if routes|default([]) %}
    routes:
{% for route in routes %}
      - route: {{ route.format(env=environment) }}
{% endfor %}
{% endif %}
...
...
{% if routes|default([]) %}
    health-check-type: http
    health-check-http-endpoint: {{ routes[0].partition('/')[1:]|join("") }}/_status?ignore-dependencies
{% endif %}
```

to be like?:
```
{% if routes|default([]) %}
    health-check-type: http
    health-check-http-endpoint: {{ routes[0].partition('/')[1:]|join("") }}/_status?ignore-dependencies
    routes:
{% for route in routes %}
      - route: {{ route.format(env=environment) }}
{% endfor %}
{% endif %}
...
...
```

idavidmcdonald:
Good points @benvand and @samuelhwilliams. I've changed to use `is defined` and also merged the two if statements into one.

PR #356, opened 2017-12-21
Add 429 metric filter and update Grafana overview dashboard

Wynndow:
We want to start rate limiting IP addresses due to notable incidents of very high traffic from particular IPs causing us to exceed the daily storage limit on our loggit stacks.

IP rate limiting at our nginx router will generate nginx access logs, which will still clutter Kibana and use storage space - exactly what we're trying to avoid. The long term plan is to _not_ send these logs to Kibana from Cloudwatch. We _will_ still want visibility on the number of 429's (rate limited responses) we're returning however. We can do this my sending a metric to Hosted Graphite and displaying it on our dashboard, as well as setting up alerting on it.

Originally this PR was going to filter the 429's from being sent to Loggit, but we decided that for the time being we would send them, so we have better visibility in Kibana into what's happening when we switch on rate limiting. Once we're happy the right things are happening, we can start filtering them out.

This PR also switches the statistic Hosted Graphite is using from `samplecount` to `sum`. The reason for this is so we can start sending metrics with a default value of zero (if, for example, we don't have any 500's within a certain period). Currently we don't send anything in these cases so Hosted Graphite uses a null value for missing metrics. Fortunately Hosted Graphite has the ability to display null values as zero (which we are currently doing), but it would be better to provide complete data. Sending default values can only really begin once we start using the `sum` statistic on our dashboards (ie. once this PR is merged and the dashboards updated). Sending it before would skew the figures horribly for graphs still using `samplecount`.

Within Hosted Graphite, we continue to use the `max` data view. This is how Hosted Graphite displays the statistic being received from CloudWatch. `max` shows the maximal value received for the metric in the time period being displayed. This is necessary due to the way Hosted Graphite handles datapoints with the same timestamp. It aggregates them all rather than only using the latest recieved; this differs to how normal Graphite works. Every minute we send the last 10 minutes worth of metric values from CloudWatch to Hosted Graphite. This means that if we used the `sum` data view we'd show massively exaggerated values.

This is how the new DM Overview dashboard would look:

screen shot 2018-01-03 at 12 00 20 2

### DIFF
--- a/grafana/dm-overview.json
+++ b/grafana/dm-overview.json
@@ -1,5 +1,5 @@
 {
-  "title": "DM Overview",
+  "title": "DM Overview with 429s",
   "tags": [],
   "style": "dark",
   "timezone": "browser",
@@ -46,7 +46,7 @@
         "current": {
           "text": "production",
           "selected": false,
-          "value": "production"
+          "value": "production",
         },
         "query": "preview,staging,production",
         "type": "custom",
@@ -74,9 +74,9 @@
   "annotations": {
     "list": []
   },
-  "refresh": "5m",
+  "refresh": "5s",
   "schemaVersion": 13,
-  "version": 2,
+  "version": 0,
   "links": [],
   "gnetId": null,
   "alertPanelMap": {},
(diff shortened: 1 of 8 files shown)


grafana/dm-overview.json
```diff
@@ -74,9 +74,9 @@
   "annotations": {
     "list": []
   },
-  "refresh": "5m",
+  "refresh": "5s",
```
idavidmcdonald: We probably don't want to change the refresh rate?

grafana/dm-overview.json
```diff
@@ -98,56 +98,56 @@
             {
               "hide": false,
               "refId": "A",
-              "target": "alias(cloudwatch.request_time_buckets.$environment.router.request_time_bucket_0.samplecount:max, '0-0.025s')"
```
idavidmcdonald: So in the past we explicitly chose (but we probably didn't do a very good job on documenting this, my bad, sorry) to use max.

This was because when leadbutt calls for a metric, it says get me metrics for the past 10 minutes. Cloudwatch returns a list of 10 timestamps with their related value.
timestamp10: 100
timestamp9: 89
...
timestamp2: 288
timestamp1: 300
where timestamp 1 is the oldest.

Then it calls again a minute later to get the data and gets a similar list but with 1 minute difference in the time period. (compare this list to the one above)
timestamp11: 49
timestamp10: 100
timestamp9: 89
...
timestamp2: 288

We used to graph based on metrics like `cloudwatch.request_time_buckets.$environment.router.request_time_bucket_0.samplecount:avg`. But we would get weird data were the value shown on the graph was half values like 170.5 which isn't a valid number of requests. It turned out that cloudwatch can give you different data for a very recent timestamp because it is still processing it. So if timestamp 1234567890 gives you value 100 when you first get it, a minute later you might get a slightly higher number like 1234567890: 117 because cloudwatch hadn't finished processing the data the first time round. So we would have the following data in HG.
1234567890: 100
1234567890: 117
And then in terms of the hostedgraphite graphing functions, using `avg` on these would give us a value of `108.5` which is incorrect. Similarly, using `sum` would give us `217` which is again wrong. By using `max` we would get the correct value. Unlike we had previously thought, the HG graphing functions are based on multiple data points for the same timestamp rather than doing an operation on a group of data covering multiple timestamps. This was the best solution we could come up with a reasonable amount of time (although retrospectively I think we should instead have looked at how are asking for metrics from cloudwatch).

So yeah, I would advise that we probably don't want to use the HG `sum` function on data coming through.

In terms of using `sum` on the cloudwatch metrics side of things, that could potentially work though so do investigate and see how you get on but you'll definitely want to try and what we get in hostedgraphite with what we are seeing logs (you could try triggering some 500s/429s yourself and confirming the number in the logs match the numbers you are getting out).

Best of luck. Oh and popping a screenshot of grafana in the PR message when you are ready for a final review would also be useful :)

Wynndow: It seems weird that Hosted Graphite works on multiple datapoints if you update a metric with the same timestamp. I've had a search around and other sources suggest that Graphite should overwrite the value for a timestamp with the latest data sent to it. I've pinged and email to their support to see if they have any insights.

Was there any reason for sending the last 10 minutes worth of metrics every minute? Other than for redundancy in case of a network error or our leadbutt app going down of something?

Wynndow: This is what I got back from Hosted Graphite. It might be worth recording this somewhere in case it crops up again in future. Might stick in in the manual.

> Hosted Graphite does work differently to normal Graphite. Graphite by default uses a last-write-wins setup so that you can overwrite your datapoints. However that also means that you have to pre-aggregate your datapoints to match the resolution of your storage scheme.

> Hosted Graphite always merges datapoints you send us for the same timestamp, in order to create our Data Views and to ensure that users don't need any kind of separate aggregation service like StatsD or carbon-aggregator. That works perfectly for real time metrics, but it does mean that updating historic datapoints with specific values isn't possible."

terraform/environments/preview/main.tf
```diff
@@ -1,5 +1,6 @@
 provider "aws" {
   region = "eu-west-1"
+  version = "1.6.0"
```
katstevens: Will we need to add this version to the staging/production terraform files?

Wynndow: ooo yes, good shout!

PR #357, opened 2017-12-22
log-streaming lambda function: substitute epochTime for timestamp in cases it seems appropriate

risicle:
This is really a workaround for nginx, which is unable to log fractional
seconds in anything other than epoch time format, which the awslogs agent
isn't able to interpret as a timestamp, so we're using this lambda function
to perform a little post-processing and re-add the precision back into the
stream.

Already deployed on preview & seems to work ok there.

### DIFF
--- a/terraform/modules/log-streaming/src/main.js
+++ b/terraform/modules/log-streaming/src/main.js
@@ -72,7 +72,20 @@ function transform(payload) {
 
         var source = buildSource(logEvent.message);
         source['instance_id'] = payload.logStream;
-        source['@timestamp'] = new Date(1 * logEvent.timestamp).toISOString();
+
+        if ("epochTime" in source && !isNaN(parseFloat(source.epochTime)) && isFinite(source.epochTime)
+            && Math.floor(Math.floor(source.epochTime) * 1000) === timestamp.getTime()) {
+            // iff epochTime is present and numeric AND ALSO represents a time "in the same second" as timestamp (i.e.
+            // timestamp could be considered a second-floored representation of epochTime) we replace timestamp with
+            // epochTime in an attempt to rescue precision for applications (*cough* nginx) which aren't able to output
+            // fractional seconds in their ISO timestamps.
+            //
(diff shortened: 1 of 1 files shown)


terraform/modules/log-streaming/src/main.js
```diff
@@ -72,7 +72,20 @@ function transform(payload) {
 
         var source = buildSource(logEvent.message);
         source['instance_id'] = payload.logStream;
-        source['@timestamp'] = new Date(1 * logEvent.timestamp).toISOString();
+
+        if ("epochTime" in source && !isNaN(parseFloat(source.epochTime)) && isFinite(source.epochTime)
+            && Math.floor(Math.floor(source.epochTime) * 1000) === timestamp.getTime()) {
```
TheDoubleK: I don't see that the outer `Math.floor` will ever do anything here, as the thing it's acting on is already guaranteed to be an integer.
e.g.
```
Math.floor(Math.floor(12345.09876) * 1000) 
= Math.floor(12345 * 1000)
= Math.floor(12345000)    // Always an integer in here so floor has no effect?
= 12345000
```
Or have I missed something?

risicle: Belt and braces. Anyone coming along to this code doesn't even need to *think* about the possibility of it being non-integer if they see that `floor`. What was js's Number weirdness around the 53bit mark again? Doesn't matter.

risicle: And also I'm just not risking something stupid going wrong in our log-critical code.

TheDoubleK: It's confusing though (at least, it confused me and I had to spend time in a Javascript console convincing myself what was going on).  Is there really a possibility that the integer multiplication inside the brackets might give a non-integer? I don't see it.

risicle: > Is there really a possibility that the integer multiplication inside the brackets might give a non-integer?

definitely not if you wrap it in `floor`

PR #359, opened 2018-01-03
Set log-metric default values to 0

Wynndow:
Now that we're using `sum` rather than `samplecount` in our Hosted
Graphite setup, we can start sending a default value for log metrics.

The default value will be set for a metric if the metric filter pattern
can't be found in any of the logs. As a note, if there are NO logs for
the period, then no metric will be recorded.

### DIFF
--- a/grafana/dm-apps.json
+++ b/grafana/dm-apps.json
@@ -426,7 +426,7 @@
           "targets": [
             {
               "refId": "A",
-              "target": "cloudwatch.application_500s.$environment.$app.500s.sum"
+              "target": "cloudwatch.application_500s.$environment.$app.500s.sum:max"
             }
           ],
           "fill": 1,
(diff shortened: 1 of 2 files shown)


grafana/dm-apps.json
```diff
@@ -426,7 +426,7 @@
           "targets": [
             {
               "refId": "A",
-              "target": "cloudwatch.application_500s.$environment.$app.500s.samplecount"
+              "target": "cloudwatch.application_500s.$environment.$app.500s.sum"
```
idavidmcdonald: So we be putting `:max` on this like the rest of our targets? I notice that it wasn't there before but not sure if that was a mistake on my behalf or intentional. Any idea?

Wynndow: Good call, it was defaulting to average without. I've added it on.

PR #364, opened 2018-01-31
Improve terraform README

idavidmcdonald:
https://trello.com/c/qhkTwalM/320-review-digitalmarketplace-aws-terraform-readmemd

Commit messages should give sufficient context (I hope)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ make requirements
 
 ## Terraform
 
-For Terraform setup and usage please check the separate ([README](terraform/README.md)).
+For Terraform setup, including non python dependencies and usage please check the separate ([README](terraform/README.md)).
 
 - [Terraform installation and setup](terraform/README.md#requirements) (requires MFA)
 - [Terraform usage](terraform/README.md#make-targets)
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -46,7 +46,7 @@ Please follow the instructions here: https://stedolan.github.io/jq/download/
 
 Follow the instructions outlined here: https://github.com/alphagov/aws-auth
 
-### Direnv (optional)
+### Direnv
 
 Install direnv (https://github.com/direnv/direnv) to automatically load environment variables from `.envrc` files
 
@@ -56,7 +56,7 @@ Install direnv (https://github.com/direnv/direnv) to automatically load environm
 
 Terraform can be used only with MFA therefore you have to set up your profiles both in `~/.aws/credentials` and `~/.aws/config`
 
-Note: You can use direnv (https://github.com/direnv/direnv) to load the AWS_PROFILE value automatically for the root modules (`aws-dm*/.envrc`). Nonetheless the Terraform wrapper script and the Makefile is going to source the `.envrc` files automatically.
+You will need to set up ```.envrc``` files for each project you wish to use which export the related ```AWS_PROFILE``` value for that project. Direnv (https://github.com/direnv/direnv) will load these automatically when using the Terraform wrapper script and Makefiles.
 
 ### Example
 
@@ -93,6 +93,14 @@ The remote state files are stored on S3 and the initialisation is automatic when
 Follow the set up guide outlined here: https://github.gds/gds/digitalmarketplace-credentials/blob/master/README.md
 
 The Makefile will look for the credentials repository at the directory defined in the DM_CREDENTIALS_REPO environment variable.
+
+You may like to add this to your `~/.bash_profile` or `.envrc` file for easy access.
+
+Example file: ~/.bash_profile
+
+```
+export DM_CREDENTIALS_REPO=/absolute/path/to/digital/marketplace/credentials/repo
+```
 
 ## Make targets
 
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -134,7 +134,7 @@ cd environments/preview
 
 ### Create an S3 bucket for storing the Terraform state files
 
-The bucket name should be: `digitalmarketplace-terraform-state-<account>`, where <account> is `main/development/production`. Make sure you turn on versioning.
+The bucket name should be: `digitalmarketplace-terraform-state-<account>`, where `<account>` is `main/development/production`. Make sure you turn on versioning.
 
 ### IAM groups / AWS users
 
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -58,6 +58,9 @@ Terraform can be used only with MFA therefore you have to set up your profiles b
 
 You will need to set up ```.envrc``` files for each project you wish to use which export the related ```AWS_PROFILE``` value for that project. Direnv (https://github.com/direnv/direnv) will load these automatically when using the Terraform wrapper script and Makefiles.
 
+You can find all the AWS profiles required in the [AWS accounts section](https://alphagov.github.io/digitalmarketplace-manual/aws-accounts.html#available-roles) of the Digital Marketplace manual.
+
+
 ### Example
 
 File: ~/.aws/credentials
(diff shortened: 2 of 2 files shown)


terraform/README.md
```diff
@@ -54,16 +67,16 @@ aws_secret_access_key=FYd4t...
 File: ~/.aws/config
 
 ```
-[profile dm-main-account-infrastructure]
+[profile dm-main-account-infrastructure-preview]
 source_profile=dm-andras
```
katstevens: Who is `dm-andras`?

terraform/README.md
```diff
@@ -68,16 +70,16 @@ aws_secret_access_key=FYd4t...
...
 
 ```
-[profile dm-main-account-infrastructure-preview]
-source_profile=dm-andras
-mfa_serial=arn:aws:iam::<main account id>:mfa/<your username>
-role_arn = arn:aws:iam::<main account id>:role/infrastructure
+[profile development-infrastructure]
+source_profile=default
+mfa_serial=arn:aws:iam::<main account id>:mfa/<IAM username>
+role_arn=arn:aws:iam::<digitalmarketplace-development account id>:role/infrastructure
```
katstevens: Aha, interesting - this might be why I haven't been able to run this stuff successfully!

terraform/README.md
```diff
@@ -1,18 +1,32 @@
...
+
+Account projects apply to the AWS account as a whole, so should only contain resources that are created once per
+account (eg root Route53 hosted zones, IAM users).
+
+Environment projects apply to individual environments and should contain resources that should be created for each
+environment (eg application logs and S3 buckets).
+
+All projects are then able to import and use the functionality they require from our folder of shared modules.
 
 ## Requirements
```
katstevens: On the top level README I think it's worth making it clear that all these bits need to be installed separately (i.e. not with pip), not just Terraform.

terraform/README.md
```diff
@@ -95,25 +119,20 @@ For more information check the Makefile contents.
...
-```
-
-This command would have the same effect:
+For example, if you wanted to run ```terraform output``` against preview you can do so:
 
 ```
-cd root/aws-dm
-. .envrc && aws-auth terraform output
+cd environments/preview
+./../../terraform-wrapper output
```
galund: This is a bit weird - are we trying to `source` the wrapper - if so, then the command needs to be
```
. ../../terraform-wrapper output
```
and if we're not trying to source it, then the initial `./` is unnecessary.

terraform/README.md
```diff
@@ -95,25 +119,20 @@ For more information check the Makefile contents.
...
 ```
 
-## Requirements in a new AWS environment
+## Requirements in a clean or completely new AWS environment
```
galund: Is the clean-ness here something about the remote environment, or something about the checkout? I don't quite follow when we are supposed to need these instructions.

terraform/README.md
galund: By installed do you mean with `make`? I'm not sure I realised that! - isn't it enough to run `make requirements` at the top level? The other makefiles are then specific to actually deploying things into AWS, rather than building local environments / build artefacts?

(I must admit I find it a bit confusing having makefiles at multiple levels, not sure I want to suggest a remedy for that though :-) )

galund:
This is looking excellent, so much more clear 👍

terraform/README.md
idavidmcdonald: So I think this is the situation where we have no existing AWS infrastructure.

E.g. we have no s3 buckets that store our terraform state, we have no users.

This is obviously not the situation that we are currently in (and unlikely to be in?).

I can either remove this section on the assumption that it will never be useful or I can give a better explanation. I'm leaning towards the later. Opinions?

terraform/README.md
idavidmcdonald: Andras is someone who set up a bunch of this (we were borrowing him from Notify briefly I think).

terraform/README.md
galund: Ah, okay yeah maybe "How we created the AWS environment".

(Trouble I suppose is: is that true? Or is this an idealised version of how we might have created it if we hadn't have been iterating something that existed already? Or is it just how we _might_ create it if we needed to - and is that a realistic thing that could happen if somehow our AWS infrastructure got wiped out by an extreme data fail scenario?)

terraform/README.md
idavidmcdonald: I've added a commit that I hope makes this a bit clearer.

terraform/README.md
idavidmcdonald: Not trying to source it. I'll drop the initial `./`

README.md
```diff
@@ -27,7 +27,7 @@ make requirements
 
 ## Terraform
 
-For Terraform setup and usage please check the separate ([README](terraform/README.md)).
+For Terraform setup, including non python dependancies and usage please check the separate ([README](terraform/README.md)).
```
samuelhwilliams: `dependencies`

terraform/README.md
```diff
@@ -1,18 +1,35 @@
...
 
 ## Requirements
 
+You will need to install the following manually (they are not installed as part of any other process such as
+`make requirements` or `pip`).
+
 ### Install Terraform
 
-Make sure you have at least v0.8 installed: https://www.terraform.io/downloads.html
+Make sure you have at least v0.11.2 installed: https://www.terraform.io/downloads.html
```
samuelhwilliams: I think this is tricky. When George and I were playing around with Terraform we noticed that it reported it was backwards incompatible (i.e. if someone has run terraform plan/apply with version v0.11.3 and I come along and try to run it with v0.11.2, it will say that I need to update my tool). This might mean that suggesting a version number here is pretty much always going to be out of date as Terraform forces everyone onto the latest version.

terraform/README.md
```diff
@@ -38,7 +55,9 @@ Install direnv (https://github.com/direnv/direnv) to automatically load environm
 
 Terraform can be used only with MFA therefore you have to set up your profiles both in `~/.aws/credentials` and `~/.aws/config`
 
-Note: You can use direnv (https://github.com/direnv/direnv) to load the AWS_PROFILE value automatically for the root modules (`aws-dm*/.envrc`). Nonetheless the Terraform wrapper script and the Makefile is going to source the `.envrc` files automatically.
+You will need to set up ```.envrc``` files for each project you wish to use which export the related ```AWS_PROFILE``` value for that project. Direnv (https://github.com/direnv/direnv) will load these automatically when using the Terraform wrapper script and Makefiles.
+
+You can find all the AWS profiles required in the AWS accounts section of the Digital Marketplace manual.
```
samuelhwilliams: We should add a link here to the relevant page.

terraform/README.md
samuelhwilliams: Might be better to link to this (https://alphagov.github.io/digitalmarketplace-manual/aws-accounts.html#available-roles) rather than provide a partial extract.

Maybe.

terraform/README.md
```diff
@@ -76,6 +95,14 @@ Follow the set up guide outlined here: https://github.gds/gds/digitalmarketplace
 
 The Makefile will look for the credentials repository at the directory defined in the DM_CREDENTIALS_REPO environment variable.
 
+You may like to add this to your `.envrc` file for easy access.
```
samuelhwilliams: Or `.bash_profile`/equivalent so that you don't need to define it in multiple places.

README.md
idavidmcdonald: Thanks. Will fix

terraform/README.md
idavidmcdonald: Will add

terraform/README.md
idavidmcdonald: I've added a paragraph to clarify this.

PR #366, opened 2018-02-16
Add metric whitelist so we only ship metrics we use

idavidmcdonald:
To save on our limit of metrics that Hosted Graphite can store,
we will only ship PaaS application metrics that we are actually
using in Hosted Graphite. This will save us lots of unused metrics
such as `requests` and `responseTime` that we are not using.

https://github.com/alphagov/paas-metric-exporter#metric-whitelist

### DIFF
--- a/vars/metric-exporter.yml
+++ b/vars/metric-exporter.yml
@@ -5,6 +5,7 @@
 API_ENDPOINT: https://api.cloud.service.gov.uk
 DEBUG: false
 METRIC_TEMPLATE: {{.Space}}.{{.App}}.{{.Instance}}.{{.Metric}}
+METRIC_WHITELIST: cpu,diskBytes,memoryBytes
 SKIP_SSL_VALIDATION: false
 UPDATE_FREQUENCY: 60
 STATSD_ENDPOINT: statsd_endpoint_to_be_replaced


katstevens:
Skimming through the metrics on HostedGraphite, I've found it hard to determine which metrics are the PaaS ones. When you refer to `responseTime` and `requests`, do you mean ones like the following, grouped under 'timers' and 'counters' (because I thought we used these?):

```
timers.preview.admin-frontend-release.0.responseTime.2xx.sum
counters.preview.admin-frontend.0.requests.2xx.count
```

Whereas the `cpu`, `diskBytes` and `memoryBytes` come from the 'gauges' section:
```
gauges.preview.admin-frontend-release.0.cpu
gauges.preview.admin-frontend-release.0.diskBytes
gauges.preview.admin-frontend-release.0.memoryBytes
```
I have no real idea what the difference is between timers, counters and gauges! Is this documented anywhere? Should it be?

idavidmcdonald:
So the rule is that if a metric is coming from the 'leadbutt app' ie. `digitalmarketplace-cloudwatch-to-graphite`, then the metric is prefixed with `cloudwatch`. I've added this fact onto my docs PR https://github.com/alphagov/digitalmarketplace-manual/pull/62.

Then there is a set of metrics sent by the Paas metrics exporter which start with a range of `timers`, `counters` and `gauges`. The prefix on these is not something we can define/control. My assumption is that they are selected based on the type of metric e.g. something that is a measurement of time is a 'timer'. I don't know if it's documented in the PaaS docs either. I don't think it's something we should add to our own docs.

A bit of confusion might be that we send some metrics that are very similar, such as count of 500s, through leadbutt and use these instead of the ones from the PaaS. However after the whitelisting of metrics is implemented, I am going to remove the PaaS metrics we didn't use from Hosted Graphite so it should be a bit easier to navigate and understand.

PR #367, opened 2018-02-19
Add script to create missing log alerts in Hosted Graphite

idavidmcdonald:
https://trello.com/c/tZghClQt/354-add-a-hosted-graphite-alert-for-missing-logs

We want to know if our logging infrastructure breaks. We can do
this by monitoring AWS `incomingLogEvents` metrics. If no logs
are being added to AWS Cloudwatch log groups for a period of time
then we can trigger an alert.

Note, we looked into setting these up as AWS Cloudwatch alerts
but ended up not going down that path. This was because
- it would be another type of infrastructure to use
- slack alerts would be extra work to set up than emails which
differs from our current alerts
- terraform doesn't let you set up cloudwatch alerts that go to
an email address (email is an unsupported protocol as described
https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html)
unless the email alert is set up manually without using terraform

Relies on metric shipping PR - https://github.com/alphagov/digitalmarketplace-cloudwatch-to-graphite/pull/7

Note, this is currently live on preview as part of my testing but not for staging/production.

I have also taken this opportunity to put into code the current 4 alerts that we have on Hosted Graphite already.

### DIFF
--- a/scripts/create-hosted-graphite-alerts.py
+++ b/scripts/create-hosted-graphite-alerts.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 """
 Description:
-    Sets up alerts for missing logs for our applications using the Hosted Graphite alerting API
+    Sets up alerts for our applications using the Hosted Graphite alerting API
 
     https://www.hostedgraphite.com/docs/alerting/alerting_api.html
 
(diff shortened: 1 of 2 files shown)


scripts/create-hosted-graphite-alerts.py
```diff
@@ -0,0 +1,94 @@
...
+    return data
+
+
+def generate_alerts(api_key):
+    endpoint = "https://api.hostedgraphite.com/v2/alerts/"
+    for environment in ENVIRONMENTS:
+        for app in APPS:
+            alert = get_alert_json(environment, app)
+            resp = requests.post(endpoint, auth=(api_key, ''), data=json.dumps(alert))
+            resp.raise_for_status()
```
Wynndow: Might be nice to have something that outputs what env/app an alert has been created for. Might help if it falls over for some reason too?

scripts/create-hosted-graphite-alerts.py
```diff
@@ -0,0 +1,94 @@
...
+        # The router app does not have application logs
+        data.update({
+            "additional_criteria": {
+                "b": {
+                    "metric": "cloudwatch.incoming_log_events.{}.{}.application_logs.sum".format(environment, app),
+                    "type": "missing",
+                    "time_period": 10,
+                }
+            },
+            "expression": "a || b",
```
katstevens: I can see `b` but not `a` - does HostedGraphite just use the top level `metric`/`alert_criteria` as 'a'?

Wynndow: I had a similar thought, found this (tldr: yes it just uses the top level): https://www.hostedgraphite.com/docs/alerting/alerting_api.html

scripts/create-hosted-graphite-alerts.py
idavidmcdonald: Will add

idavidmcdonald:
I've looked to set up the rest of our alerts using this script. During that I've uncovered a bug with the hosted graphite API. I'm in contact with their support team and they say they are going to fix it (no idea on time frame but sounded like in the short term).

I've got a commit waiting on my local to do it but until they fix their API it will have a slightly broken interface for our alerts so I'll hold off pushing it up for the moment...

scripts/create-hosted-graphite-alerts.py
```diff
@@ -0,0 +1,95 @@
...
+import requests
+from docopt import docopt
+
+
+# No staging alert as we have a limit on how many alerts we can have and this will cover both the first step of our
+# pipeline and also production
+ENVIRONMENTS = ["preview", "production"]
+
+APPS = ["api", "search-api", "admin-frontend", "buyer-frontend", "briefs-frontend", "brief-responses-frontend",
+        "router", "supplier-frontend", "user-frontend"]
```
risicle: Do we not have an existing list of "all apps" we can piggyback on here? I can imagine someone adding a new app (if I grit my teeth :grimacing:) and neglecting to add it here...

scripts/create-hosted-graphite-alerts.py
```diff
@@ -0,0 +1,95 @@
...
+        "notification_channels": ["Notify DM 2ndline"],  # Hardcoded name, channel had been set up manually already
+        "notification_type": ["every", 60],
+        "info": """No incoming log events metrics for the last 10 minutes for the {} app. This could be either the \
+application logs or the nginx logs or both. This could indicate either a problem with metric shipping to Hosted \
+Graphite or that the logs are not being created.\nDO NOT MANUALLY EDIT - Set up through Hosted Graphite API so GUI may \
+have inconsistencies. See HG alerting API for details""".format(app)
+    }
+
+    if app != "router":
+        # The router app does not have application logs
```
risicle: Yet

scripts/create-hosted-graphite-alerts.py
CrystalPea: Ahh, fair point!

risicle:
Funny you should say that... think it's just managed to do that for itself...

scripts/create-hosted-graphite-alerts.py
idavidmcdonald: Good question, let me look into this

idavidmcdonald:
So @risicle I looked into the list of app names.

We do have some similar but different lists that already exist:
- `vars/common.yml` contains all the apps but not as a list and there are extra keys that we would not want such as `db-backup` and `maintenance_mode`
- some of our terraform modules do define them, for example `terraform/modules/application-logs/variables.tf`.  However this one doesn't include the router app for example and also isn't some nice python file to import.

So it looks like there isn't a list we could just use as is. This is the first time we are using this exact list I think. As this is the first time I'm probably not going to extract out the list and put it somewhere common although I can do if you feel strongly (if so, feel free to make a recommendation of where it could go if you have one).

risicle:
No I don't feel strongly about it - it might be an interesting thing to do, having a more central place where the different categories of application are listed & defined...

PR #368, opened 2018-02-22
Update db migrate definition

samuelhwilliams:
Update makefile to use api venv now that we have an updated docker image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
 	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none -i 1 -m 128M -c 'sleep 2h'
 	cf stop ${APPLICATION_NAME}-db-migration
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -91,7 +91,7 @@ deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
 	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none -i 1 -m 128M -c 'sleep 2h'
 	cf stop ${APPLICATION_NAME}-db-migration
-	cf run-task ${APPLICATION_NAME}-db-migration "cd /app && python application.py db upgrade" --name ${APPLICATION_NAME}-db-migration
+	cf run-task ${APPLICATION_NAME}-db-migration "cd /app && venv/bin/python application.py db upgrade" --name ${APPLICATION_NAME}-db-migration
```
brenetic: `VIRTUALENV_ROOT` ?

samuelhwilliams: That won't be defined at this point, I think, because the venv hasn't been activated. That var gets set once you activate it.

samuelhwilliams: Maybe not?

katstevens: Can we call the `virtualenv` step at the beginning of this step (as per some of the other steps, e.g. `make generate-manifest`)?

PR #370, opened 2018-03-05
default.nix: bump terraform to 0.11.3

risicle:
Please people, stop blindly bumping your terraform versions. This is getting tiring. Can we settle on 0.11.2 for a while maybe?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -18,12 +18,12 @@ in (with args; {
       pkgs.aws-auth
       pkgs.sops
       (pkgs.terraform.overrideAttrs (oldAttrs: rec {
-        name = "terraform-0.11.1";
+        name = "terraform-0.11.3";
         src = pkgs.fetchFromGitHub {
           owner  = "hashicorp";
           repo   = "terraform";
-          rev    = "v0.11.1";
-          sha256 = "04qyhlif3b3kjs3m6c3mx45sgr5r13x55aic638zzlrhbpmqiih1";
+          rev    = "v0.11.3";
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Are we 100% sure that there isn't someone who already has 0.11.3 installed, which will make 0.11.2 redundant as soon as they do something on the 2nd line rota?

The `terraform/README.md` currently says "install the latest version", so if we're going to pin it at 0.11.2 for a while we should probably update that (weirdly nested) README to reflect that at the same time.

Wynndow:
Be worth pinning it in the terraform too: https://www.terraform.io/docs/configuration/terraform.html

samuelhwilliams:
Probably easier to pin it at the current latest, 0.11.3, in case as @TheDoubleK says, someone's already got that installed.

Wynndow:
Yeah, I've got 0.11.3, and that's what was used on the main account on Monday.

risicle:
Well let's be real - it's `default.nix`, this is just for me...

risicle:
I've upgraded it to 0.11.3, but I don't really get the point of pinning it in terraform's config...

Wynndow:
It should prevent someone from accidentally advancing the Terraform version for everyone else?

risicle:
Oh I see, it can specify a *maximum* version, too.

Issue I can see with this though - for the rest of you is it actually easy to have an "arbitrary" terraform version installed? What do you use? Homebrew?

Wynndow:
Yeah, homebrew

PR #371, opened 2018-03-05
Add Jenkins_2

Wynndow:
This adds the Terraform code to spin up a new instance for a new Jenkins
box. This is being done to allow pen testers to test a Jenkins instance
that isn't our actual instance. It will be provisioned via Ansiible to
be a similar as we can get it.

This Terraform code uses the existing security group that is applied to
our current box, and the same instance profile.

There is an extra security group applied, which is to allow access to
the box from Cyberis's IP's.

This could lay the groundwork for a new, upgraded Jenkins box which
could be built in the future. Alternatively, this Terraform could be
scrapped and we could move Jenkins to the PaaS.

### DIFF
--- a/terraform/accounts/development/main.tf
+++ b/terraform/accounts/development/main.tf
@@ -1,5 +1,6 @@
 provider "aws" {
   region = "eu-west-1"
+  version = "1.9.0"
 }
 
 terraform {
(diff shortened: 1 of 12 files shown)


brenetic:
👍 for the provider version addition

PR #373, opened 2018-03-13
Only drop BriefResponses during DB clean if procurement is ongoing

katstevens:
Currently all BriefResponses are dropped during the DB sanitisation, to prevent identifying suppliers involved in an ongoing procurement.

However if the BriefResponse has an 'awarded' status, the procurement has by definition ended. It's useful to have this award data in the dev environments so don't delete it during the DB clean.

### DIFF
--- a/scripts/clean-db-dump.sql
+++ b/scripts/clean-db-dump.sql
@@ -25,10 +25,17 @@ WHERE framework_id IN
      WHERE status='open');
 
 -- Remove data related to open DOS procurements
--- (We can't tell if a procurement is ongoing, so delete all brief responses)
+-- (Delete all brief responses except those with 'awarded' status, where the procurement has ended)
 \echo 'Delete brief responses'
 DELETE
-FROM brief_responses;
+FROM brief_responses
+WHERE awarded_at IS NOT NULL;
+
+-- Sanitise personal data in awarded brief responses
+\echo 'Update brief responses'
+UPDATE brief_responses
+SET data = data::jsonb || '{"respondToEmailAddress": "cleaned-example-email@gov.uk"}'::jsonb
+WHERE data::TEXT != '{}';
 
 \echo 'Delete draft briefs'
 \echo '  > Delete draft brief users'


galund:
(Added Ben - I don't think there's any PID in the brief response data, so I don't think there are GDPR implications, but FYI anyway.)

katstevens:
Good shout - as well as the public supplier information (`supplierName`, `supplierOrganisationSize`) there's also the buyer `respondToEmailAddress` - that one may require sanitisation?

benvand:
Yeah, we will need to obfuscate the `"respondToEmailAddress"` in the data blob. Also `brief_responses` doesn't have a status.
One way of doing this would be:
```UPDATE brief_responses SET data = jsonb_set(data::jsonb, '{responsdToEmailAddress}', '"simulate-delivered@notifications.service.gov.uk"');
delete from brief_responses where brief_id not in (select id from briefs where awarded_at is not null)```

**Comment by katstevens** [comment]
Ah the status is isn't a proper DB field, is it. Ugh. Cool, thanks for the suggestion Ben!

**Comment by katstevens** [comment]
I've tried out some fancy JSON updating - seems to work locally in `psql` (Postgres v9.5.7).

File: scripts/clean-db-dump.sql [hunk 1]
```diff
@@ -25,10 +25,18 @@ WHERE framework_id IN
...
-FROM brief_responses;
+FROM brief_responses
+WHERE awarded_at IS NOT NULL;
+
+-- Sanitise personal data in awarded brief responses
+\echo 'Update brief responses'
+UPDATE brief_responses
+SET data = data::jsonb - 'respondToEmailAddress' || '{"respondToEmailAddress": "cleaned-example-email@gov.uk"}'::jsonb
+WHERE data IS NOT NULL
+AND cast(data AS TEXT) != 'null';
```
galund: Presumably this is _necessary_ in order for the update to not fail -- because if it isn't required to be there then I imagine it's just slowing the query down.

katstevens: I copied it from another JSON blob sanitisation below - `signed_agreement_details`.

benvand: You shouldn't need the `- 'respondToEmailAddress'` the `||` updates existing.

There are 2 different ways of casting used there, the `::JSONB` and the `cast(data AS TEXT)`.

Effectively:
`cast(data AS TEXT) == data::TEXT`
and
`cast(data AS JSONB) == data::jsonb`

We may also need to limit on `data::text != '{}'` IIRC 🤔

katstevens: That makes sense. I'll try and apply the casting consistently. Do you know if we need both `data IS NOT NULL AND cast(data AS TEXT) != 'null'`?

benvand: Actually I don't _think_ we need either of those.
Our json fields should be `'{}'::json` if empty.

PR #374, opened 2018-03-20
Allow disabling MFA

samuelhwilliams:
## Summary
I have a new phone and I can't change my MFA in the console as it needs
me to disable my existing MFA before I can, presumably, setup my new
one. So everyone needs permission to disable their own MFA.


MuriloDalRi:
You're just trying to show off your new phone

PR #381, opened 2018-04-05
Update metric alert timing from 10min to 15min

katstevens:
Changes the missing log alert threshold from 10 minutes to 15 minutes.

While looking at this, I realised a) I didn't fully understand the Hosted Graphite scripts b) the only way to test them was to actually run them for real. So I wrote some tests! This meant moving most of their code into new modules within the `dmaws` folder (the top level scripts remain where they are, so the paths won't have changed).

Also there was no `make requirements-dev` command for this repo, so I added one 👼

### DIFF
--- a/dmaws/hosted_graphite/create_alerts.py
+++ b/dmaws/hosted_graphite/create_alerts.py
@@ -104,23 +104,26 @@ have inconsistencies. See HG alerting API for details""".format(app)
     return data
 
 
-def create_missing_logs_alerts(api_key):
-    # No staging alert as we have a limit on how many alerts we can have and this will cover both the first step of our
-    # pipeline and also production
-    ENVIRONMENTS = ["preview", "production"]
-    APPS = ["api", "search-api", "admin-frontend", "buyer-frontend", "briefs-frontend", "brief-responses-frontend",
-            "router", "supplier-frontend", "user-frontend"]
-
-    for environment in ENVIRONMENTS:
-        for app in APPS:
-            print("Creating missing logs alert for {} {}".format(environment, app))
-            create_alert(api_key, get_missing_logs_alert_json(environment, app))
-
-
 def create_alert(api_key, alert):
     endpoint = "https://api.hostedgraphite.com/v2/alerts/"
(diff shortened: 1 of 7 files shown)


dmaws/hosted_graphite/create_alerts.py
```diff
@@ -0,0 +1,134 @@
...
+
+    `Our UI doesn’t fully support composites at the moment. As a result, composite alerts cannot be edited via the UI -
+    it needs to be done via the API. The alert overview page (when you click the eye button on an alert) will only
+    display one metric for the alert instead of all the metrics associated. However the alert notifications are working
+    and will display the graph of the last metric that breached the alert threshold. So for example, if the alert is
+    a && b, and a breaches the threshold, then a few minutes later b breaches it’s threshold, the alert notification
+    will show the metric graph for b`
+
+    Time period for alerting missing logs is set to 15 minutes. The smoke tests run every 5 minutes, this allows
+    some wiggle room to avoid false positives arising from a delay in shipping the metrics from Cloudwatch.
```
idavidmcdonald: It feels like the part about avoiding false positives should be linked to the sentence about the time period being 15 minutes rather than linked with the 5 minute smoke test sentence

dmaws/hosted_graphite/create_alerts.py
```diff
@@ -0,0 +1,134 @@
...
+
+def create_alert(api_key, alert):
+    endpoint = "https://api.hostedgraphite.com/v2/alerts/"
+    resp = requests.post(endpoint, auth=(api_key, ''), data=json.dumps(alert))
+    resp.raise_for_status()
+
+
+# No staging alert as we have a limit on how many alerts we can have and this will cover both the first step of our
+# pipeline and also production
+ALERT_ENVIRONMENTS = ["preview", "production"]
```
idavidmcdonald: It feels like this and the variable below it should be in the script rather than this library file (would involve probably passing it as a param to the create_log functions). Aware that is a bit of work though so can go either way...

katstevens: Agreed. It's not too much work and it'll be useful if we want to eventually run the script for just one alert.

PR #384, opened 2018-04-10
Move maintenance mode variable

samuelhwilliams:
## Summary
Rather than defining maintenance mode in the common vars, which will
make it affect all environments simultaneously, we should define it on a
per-environment basis so that we are able to toggle it for specific
environments without the risk of accidentally also turning it on for
production.

This will allow us to more safely have a job that can toggle maintenance
mode on/off.

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -1,6 +1,4 @@
 ---
-maintenance_mode: live
-
 router:
   routes:
     - www.{env}.marketplace.team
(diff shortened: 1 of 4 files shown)


katstevens:
The line in the docs will need updating: https://alphagov.github.io/digitalmarketplace-manual/maintenance-mode.html?highlight=maintenance#activate (we were going to update it anyway to fix the `RELEASE_NAME`, weren't we?)

samuelhwilliams:
Yep, I'll update all that jazz shortly.

PR #388, opened 2018-04-19
Enforce formatting on terraform files

samuelhwilliams:
## Summary
Start enforcing consistent formatting of Terraform configuration files through
Travis with a new `Makefile` step that runs `terraform fmt` on Pull Requests.
Also applies an initial sweep of re-formatting, which is mostly about aligning
variable values with correct spacing.

## Ticket
https://trello.com/c/ubqlb6pm/412-enforce-terraform-code-style

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,11 @@ python:
   - "2.7"
   - "3.6"
 env:
-  - TERRAFORM_VERSION=0.11.7 TERRAFORM_DOWNLOAD_URL=https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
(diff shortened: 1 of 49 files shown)


risicle:
That does look like an *extremely* annoying format to edit in.

risicle:
(though the test does pass for me & work on my setup)

samuelhwilliams:
Why does it look annoying? You can basically edit it however you want and then just run the formatter.

Although honestly I prefer editing files like this and manually aligning the things /shrug

risicle:
Does your space bar tend to wear out early? Personally I like editing content and not whitespace. "Oh look I've just made this key longer than any of the other keys in its paragraph, guess I'm going to be shifting the equals sign five to the right on all these lines too" - and then you end up with buckets of diff spam cluttering up your PRs and your `git blame`.

Is there even a choice over formatting settings? Because otherwise this is an academic discussion.

samuelhwilliams:
This appears to be the standard format, which is what GOV.UK/PaaS use, so I think for cross-GDS consistency we'd just keep this.

risicle:
fair enough

samuelhwilliams:
FYI If you don't want to edit in the format this enforces, you can just do whatever you want and then run `terraform fmt -write=true -diff=true -list=true terraform` and it will fix everything for you. I will add this as a makefile step under `make terraformat-apply`

PR #393, opened 2018-05-20
Inject callback token in environment

samuelhwilliams:
## Summary
We are adding a new route/blueprint to the API for Notify to use to post
the successful/failed delivery of a notification. Given it's a new
token, we need to start injecting it into the API's environment.

 ## Ticket
https://trello.com/c/vOAUDwnL/446

### DIFF
--- a/paas/api.j2
+++ b/paas/api.j2
@@ -7,6 +7,8 @@
 
       DM_API_AUTH_TOKENS: {{ api.auth_tokens|join(':') }}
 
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Yeah I've been thinking about that too. I think it's probably reasonable to just stick it in that job as well. 👍

samuelhwilliams:
https://github.com/alphagov/digitalmarketplace-scripts/pull/219

PR #402, opened 2018-06-06
Add new backups account with s3 bucket

benvand:
Requires:
https://github.com/alphagov/digitalmarketplace-credentials/pull/117

Add the new backups account with an s3 bucket with a 6 month lifecycle policy


terraform/modules/cloudtrail/cloudtrail-bucket/main.tf
```diff
@@ -0,0 +1,35 @@
+resource "aws_s3_bucket" "cloudtrail_bucket" {
+  bucket        = "${var.s3_bucket_name}"
```
Wynndow: Could this have a more descriptive variable name? `cloudtrail_s3_bucket_name` maybe.

terraform/modules/cloudtrail/cloudtrail-cloudwatch/variables.tf
```diff
@@ -0,0 +1,8 @@
+variable "trail_name" {
+  description = "The name of the CloudTrail trail we are setting up CloudWatch for."
+}
+
+variable "retention_in_days" {
+  default     = 730
```
Wynndow: We're using 731 above, maybe this should match.

terraform/accounts/backups/access.tf
```diff
@@ -0,0 +1,74 @@
...
+  name = "backups-access-policy"
+
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:ListAllMyBuckets",
```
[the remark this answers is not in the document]
benvand: Good call that shouldn't have made it in

terraform/modules/cloudtrail/cloudtrail/outputs.tf
```diff
@@ -0,0 +1 @@
+
```
Wynndow: Can this be deleted if we're not outputting anything?

terraform/accounts/main/main.tf
```diff
@@ -22,6 +26,7 @@ module "iam_common" {
 module "iam_users" {
   source                          = "../../modules/iam-users"
   admins                          = "${var.admins}"
+  backups                         = "${var.admins}"                                                // Add admins to backups group, gives 2nd line access to backups
```
Wynndow: Wasn't this going to be the same as `prod_infrastructure_users`?

terraform/accounts/backups/access.tf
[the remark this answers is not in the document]
benvand: 👍 Done

terraform/modules/cloudtrail/cloudtrail-bucket/main.tf
benvand: This variable name mirrors the variable name given by terraform here:
https://www.terraform.io/docs/providers/aws/r/cloudtrail.html#s3_bucket_name

The choices are I guess:
1) To change it globally in our code, make every instance of `s3_bucket_name` into `cloudtrail_s3_bucket_name`
2) Only change it in the `cloudtrail/cloudtrail-bucket` submodule
3) Leave it as is
I'm fairly easy. I think I'd prefer 1 or 3, that way the variable has a globally consistent name on our side. It's just whether there's more value in having that name be the same as the cloudtrail terraform variable or have it be more descriptive.

Any preference?

terraform/modules/cloudtrail/cloudtrail-cloudwatch/variables.tf
benvand: Good call, 730 won't run, it's not a valid choice iirc

terraform/modules/cloudtrail/cloudtrail/outputs.tf
benvand: Yeah I added this because of this bit in the docs:
https://www.terraform.io/docs/modules/create.html#standard-module-structure
>>> main.tf, variables.tf, outputs.tf. These are the recommended filenames for a minimal module, even if they're empty. main.tf should be the primary entrypoint. For a simple module, this may be where all the resources are created. For a complex module, resource creation may be split into multiple files but all nested module usage should be in the main file. variables.tf and outputs.tf should contain the declarations for variables and outputs, respectively.

But it's not consistent with the rest of our modules and it doesn't do anything so actually I'm going to get rid. If we're going to have empty files just to indicate no input and no output then we should do it globally.

terraform/accounts/main/main.tf
benvand: Good call, admins can assume anything anyway (tested) 👍

terraform/modules/cloudtrail/cloudtrail-bucket/main.tf
Wynndow: Go with 3, it's not a big issue. It's meaning gets clarified by the context it's being used in anyway.

terraform/modules/cloudtrail/cloudtrail/outputs.tf
Wynndow: Interesting - wasn't aware of those recommendations. But yeah, agree consistency is probably better otherwise the empty files existence may be confusing.

terraform/accounts/main/main.tf
Wynndow: Nice. Good to know.

benvand:
Thanks @Wynndow !!

benvand:
I ❤️ terraform now

PR #408, opened 2018-07-09
S3 buckets

benvand:
Terraform for preview staging and production s3 buckets.

See [README.md](https://github.com/alphagov/digitalmarketplace-aws/pull/408/files#diff-e3ac418a4fe3b69c3e62825d6d1001d1) for more info

### DIFF
--- a/terraform/environments/preview/s3_buckets.tf
+++ b/terraform/environments/preview/s3_buckets.tf
@@ -1,0 +1,168 @@
+# Other buckets should be set to log to this bucket
+resource "aws_s3_bucket" "server_access_logs_bucket" {
+  bucket = "digitalmarketplace-logs-preview-preview"
+  acl    = "log-delivery-write"
+
+  versioning {
+    enabled = true
+  }
+}
+
+# TODO remove these hard-coded definitions in favour of using the terraform/modules/s3-document-bucket module after the
+# tf v0.12 upgrade. We need to contitionally include the principals block
+
+# Agreements - devs: read write list
+
+data "aws_iam_policy_document" "agreements_bucket_policy_document" {
+  statement {
+    effect = "Allow"
+
+    principals {
+      identifiers = ["arn:aws:iam::${var.aws_dev_account_id}:role/developers"]
+      type        = "AWS"
+    }
+
+    actions = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:PutObjectAcl",
+      "s3:GetBucketLocation",
+      "s3:ListBucket",
+    ]
+
+    resources = [
+      "arn:aws:s3:::digitalmarketplace-agreements-preview-preview/*",
+      "arn:aws:s3:::digitalmarketplace-agreements-preview-preview",
+    ]
+  }
+}
+
+resource "aws_s3_bucket" "agreements_bucket" {
+  bucket = "digitalmarketplace-agreements-preview-preview"
+  acl    = "private"
+
+  versioning {
+    enabled = true
+  }
+
+  logging {
+    target_bucket = "${aws_s3_bucket.server_access_logs_bucket.id}"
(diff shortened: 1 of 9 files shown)


katstevens:
Can we put a line in the S3 section of the manual (https://alphagov.github.io/digitalmarketplace-manual/aws-resources.html#s3-buckets) to say which role you need to switch to for each environment, in order to view the buckets defined here? I was double checking what we had, and before I'd switched role I could view a whole bunch of old buckets (and the public bucket, which we definitely need for at least one Jenkins script) that I don't think are defined anywhere in this repo.

benvand:
@katstevens 
> Do we need to explicitly configure ACL as part of the policy, or should we update the Cloudtrail stuff to match this?

Yeah we should update the old style definition to use computed IAM policy document data sources. It doesn't actually make any difference to the end product but the policy document style is way more readable.

> Can we put a line in the S3 section of the manual (https://alphagov.github.io/digitalmarketplace-manual/aws-resources.html#s3-buckets) to say which role you need to switch to for each environment, in order to view the buckets defined here?

Sure, we'll need to keep this up to date when we change them though. I'll see if I can link to the relevant groups in creds 🤷‍♀️

benvand:
Also re the old buckets, yeah, they're in the main account, and probably shouldn't be, but they're out of scope of this PR (But we should do something about them soon.)

terraform/environments/preview/s3_buckets.tf
```diff
@@ -0,0 +1,168 @@
...
+    target_bucket = "${aws_s3_bucket.server_access_logs_bucket.id}"
+    target_prefix = "digitalmarketplace-documents-preview-preview/"
+  }
+
+  policy = "${data.aws_iam_policy_document.documents_bucket_policy_document.json}"
+}
+
+# G7-draft-documents
+
+resource "aws_s3_bucket" "g7-draft-documents_bucket" {
```
Wynndow: No policy for this bucket?

terraform/environments/preview/s3_buckets.tf
```diff
@@ -0,0 +1,168 @@
...
+
+  logging {
+    target_bucket = "${aws_s3_bucket.server_access_logs_bucket.id}"
+    target_prefix = "digitalmarketplace-g7-draft-documents-preview-preview/"
+  }
+}
+
+# Submissions
+
+resource "aws_s3_bucket" "submissions_bucket" {
```
Wynndow: No policy on this bucket either?

terraform/environments/preview/s3_buckets.tf
```diff
@@ -0,0 +1,168 @@
...
+  versioning {
+    enabled = true
+  }
+
+  logging {
+    target_bucket = "${aws_s3_bucket.server_access_logs_bucket.id}"
+    target_prefix = "digitalmarketplace-agreements-preview-preview/"
+  }
+
+  policy = "${data.aws_iam_policy_document.agreements_bucket_policy_document.json}"
```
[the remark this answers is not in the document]
benvand: Yeah, pretty nifty. Better than writing plain json

terraform/environments/preview/s3_buckets.tf
benvand: screen shot 2018-07-26 at 15 36 41

benvand: screen shot 2018-07-26 at 15 36 58

benvand: Nada 🤔  Bit weird

terraform/environments/preview/s3_buckets.tf
benvand: screen shot 2018-07-26 at 15 36 41

Same

PR #412, opened 2018-07-23
Add retries to grafana overview dashboard

Wynndow:
### Fix bugs with grafana dashboard script

docopt needs to be imported from docopt, otherwise you get a `module not
callable` error.

The path to the json files also needed an extra `..`

### Add retries to overview dashboard

Adds a new panel to the overview dashboard which aggregates the values of
the retry metric for all the apps and displays it.

This is already applied, if you look at the monitor you should see how it looks.


Wynndow:
yeah!

PR #414, opened 2018-07-27
Jenkins2 becomes Jenkins3. And some tweaks.

Wynndow:
Change jenkins2 to jenkins3. This is due to running into the letsencrypt rate limit for certificates of 20 per week. Using a different subdomain allowed me to continue developing. After new Jenkins is out we can point the original domain at the new box so it shoudn't be too much of an issue.

This also adds a security group file to the Jenkins module - this is mostly just a straight copy of the jenkins security account defined in the main account (which we can drop after the move). The main difference is the programatic inclusion of the jenkins elastic ip address in the inbound https rule. This just means we don't need to update the terraform credentials file with the new ip address.

The new ec2 instance definition programatically pulls in the new security group rather than hard coding the id of the existing security group. Better.

### DIFF
--- a/terraform/accounts/main/main.tf
+++ b/terraform/accounts/main/main.tf
@@ -55,10 +55,10 @@ module "sops_credentials" {
 }
 
 module "jenkins" {
-  source                     = "../../modules/jenkins"
-  aws_main_account_id        = "${var.aws_main_account_id}"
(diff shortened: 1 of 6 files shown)


Wynndow:
Yep, that's right.

PR #417, opened 2018-08-01
Put Jenkins3 behind ELB

Wynndow:
This was done to get a free certificate, rather than having to get one
from letsencrypt. The main motivation was I kept bumping in to the
letsencrypt rate limit on issuing certificates of 20 per week.

This code sets up an CNAME record for ci3.market.... which points to the
DNS name of the elb. The elb terminates our tls and proxies the request
on to the Jenkins instance via http. This is all within our VPC and
subnet.

The ELB also passes tcp connections through port 22 to allow ssh access.

There are two security groups - one for the ELB with all our usual
whitelisted IP address as the only allowed ingress, and one for the
instance itself, which only allowed ingress from the ELB.

### DIFF
--- a/terraform/modules/jenkins/elb.tf
+++ b/terraform/modules/jenkins/elb.tf
@@ -1,0 +1,64 @@
+resource "aws_elb" "jenkins_elb" {
+  name            = "Jenkins-ELB"
+  subnets         = ["${aws_instance.jenkins3.subnet_id}"]
+  instances       = ["${aws_instance.jenkins3.id}"]
+  security_groups = ["${aws_security_group.jenkins_elb_security_group.id}"]
+
+  listener {
+    instance_port      = 80
+    instance_protocol  = "http"
+    lb_port            = 443
+    lb_protocol        = "https"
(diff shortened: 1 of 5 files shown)


terraform/modules/jenkins/elb.tf
```diff
@@ -0,0 +1,64 @@
...
+    instance_protocol = "tcp"
+    lb_port           = 22
+    lb_protocol       = "tcp"
+  }
+
+  health_check {
+    healthy_threshold   = 2
+    unhealthy_threshold = 2
+    timeout             = 5
+    target              = "TCP:22"
```
risicle: I guess the thought is that as long as we have SSH access we can take a peek at what's going wrong the the web interface, but not having SSH we're stuffed and need a reboot?

Wynndow: It was more than when bringing up a new, clean instance, nothing is listening on port 80 as nginx isn't there, so there'll be no 200 OK. ssh is there though.

PR #419, opened 2018-08-09
Antivirus SNS notifications

Wynndow:
For [this trello ticket](https://trello.com/c/DwwT9fw5).

### Add antivirus-sns module  …
The module sets up an SNS topic for sending notifications to the
antivirus app.

### Add S3 notifications and format  …
We need to set up notifications for the S3 buckets we want to monitor
for file uploads. These notifications will publish a message to the file
upload topic whenever a file is created in any of them. That published
message is then distributed to all subscriptions. That's next.

### Add sns topic subscription  …
This adds a subscription to the file upload topic for an endpoint on the
antivirus api. The api will be using basic auth which is included in the
endpoint described here. The creds are in the creds repo.

The endpoint _has_ to be able to autoconfirm the subscription request
that is sent to it or this code will not apply.

### DIFF
--- a/terraform/modules/antivirus-sns/subscriptions.tf
+++ b/terraform/modules/antivirus-sns/subscriptions.tf
@@ -1,0 +1,6 @@
+resource "aws_sns_topic_subscription" "antivirus_to_s3_file_upload_notification_topic" {
+  topic_arn              = "${aws_sns_topic.s3_file_upload_notification.arn}"
+  protocol               = "https"
+  endpoint               = "https://${var.antivirus_api_basic_auth}@${var.antivirus_api_host}/sns/s3/uploaded"
+  endpoint_auto_confirms = "true"
+}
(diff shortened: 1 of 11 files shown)


terraform/modules/antivirus-sns/subscriptions.tf
```diff
@@ -0,0 +1,6 @@
+resource "aws_sns_topic_subscription" "antivirus_to_s3_file_upload_notification_topic" {
+  topic_arn              = "${aws_sns_topic.s3_file_upload_notification.arn}"
+  protocol               = "https"
+  endpoint               = "https://${var.antivirus_api_basic_auth}@${var.antivirus_api_url}/callback"
```
risicle: As I said, miiiight be nicer to have this as `antivirus_api_host` rather than `antivirus_api_url` as it's clearer and just accept we're going to handle all the path stuff here.

PR #425, opened 2018-08-23
Revert "Scale supplier and api for DOS3 applications"

Wynndow:
This reverts commit 35c1b56e3242ec6b563ea04c0e10e301b9753499.

DOS3 applications are now closed so we can return to the default
instance count and memory size of the api and supplier fe.


lfdebrux:
This makes sense to me

PR #426, opened 2018-08-23
Refactor Jenkins module

Wynndow:
Whilst writing docs on how to spin up new Jenkins in case of an emergency,
or just to do an upgrade similar to what we just did, I realised that the 
Jenkins module was not modular enough, and it was not possible to bring
up another module. This should fix that. ~~Some of the existing terraform state
will need moving about so it matches what's been moved in/out of the module
but that should be fine.~~ I've moved the state in/out of the module and it all
works perfectly.

The change actually makes _no_ change to our infrastructure, it's just a refactor
of the Terraform.

This refactors the Jenkins module to make it actually reusable.
Previously it defined resources within the module that would not be
duplicatable had an extra module been created.

The Jenkins instance profile has been moved out of the module and into
the main account. We only want one of these - quite a few things depend
on knowing the exact arn of the jenkins instance profile so it doesn't
make sense to create a new one.

The A record for Jenkins has been moved out of the route53 file in the
main account and into the module. This wasn't strictly neccessary, but
it keeps all the changeable definitions for the module in one place.

The Jenkins EC2 instance is now defined with variables that are passed
in so different modules can use different values for ami and instance
type. Useful if upgrading.

The wildcard certificate for Jenkins ELB has moved out of the module and
into the main account. It's a wildcard, so we can't/don't need to define
it more than once for different modules/subdomains.


terraform/modules/jenkins/ec2.tf
```diff
@@ -3,14 +3,14 @@ provider "aws" {
 }
 
 resource "aws_instance" "jenkins3" {
```
lfdebrux: Does this still need to be called jenkins3?

Wynndow: Possibly not actually. I previously thought it wasn't going to be able to change these names, but have since learned that it might. I'll have a go.

Wynndow: Yep, it was totally possible. All done.

lfdebrux: Nice!

Wynndow:
Perfect! Thanks Laurence

PR #427, opened 2018-08-24
Mute output from cf set-env to hide credentials

katstevens:
Trello https://trello.com/c/eBl3yUUT/27-jenkins-credentials-shown-in-plain-text-in-job-scripts-and-logs#

NB: not tested yet (to avoid strain on prod API)

See Jenkins database-backup output.


lfdebrux:
As we weren't able to test this today we should check the output of the console log for this job once it has run after merging, just to make sure it did what we wanted!

katstevens:
Thinking about it, is there a reason why we use `cf set-env` here instead of setting the env variables in the manifest? We're using python scripts to generate the post data and to generate the manifest, why not put them together in Python? (Then we could... write a test for it? 🙀)

Wynndow:
Yeah! That's a much better way!

PR #429, opened 2018-08-29
98-set-up-cross-region-replication-for-our-backups-s3-buckets

benvand:
#### Setting up cross region replication of incoming backups.

Predominantly followed this example, changing names to be in line with existing convention:
https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#using-replication-configuration

Only new incoming backups will be replicated (as per https://docs.aws.amazon.com/AmazonS3/latest/dev/crr-what-is-isnot-replicated.html) and the backups will be kept for 7 days (see https://github.com/alphagov/digitalmarketplace-aws/pull/429/files#diff-fe12ef74960acd2feb5dc4ef2c4d6a66R75); this has been cleared with the dev team and @louzoid-gds .

In order to give the new region bucket similar permissions to the regular backups bucket I removed the hard coded json bucket policy and created 2 policy documents. One inheriting from the other. 
`aws_iam_policy_document.cross_region_database_backups_s3_bucket_policy_document`
https://github.com/alphagov/digitalmarketplace-aws/pull/429/files#diff-fe12ef74960acd2feb5dc4ef2c4d6a66R2
and `aws_iam_policy_document.database_backups_s3_bucket_policy_document`
https://github.com/alphagov/digitalmarketplace-aws/pull/429/files#diff-fe12ef74960acd2feb5dc4ef2c4d6a66R48

We also need to apply `bucket-owner-read` acl to uploaded db dumps
    
As per https://docs.aws.amazon.com/AmazonS3/latest/dev/crr-troubleshoot.html

> By default, the bucket owner does not have any permissions on the objects created by <another> account. And the replication configuration replicates only the objects for which the bucket owner has access permissions.

Therefore the bucket owner must be granted read permissions on the uploaded object for replication to work.
We do this by applying the `bucket-owner-read` canned acl detailed here:

> Object owner gets FULL_CONTROL. Bucket owner gets READ access.

https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl

### DIFF
--- a/terraform/accounts/backups/s3_buckets.tf
+++ b/terraform/accounts/backups/s3_buckets.tf
@@ -1,3 +1,47 @@
+resource "aws_s3_bucket" "cross_region_database_backups_s3_bucket" {
+  provider = "aws.london"
+  bucket   = "digitalmarketplace-cross-region-database-backups"
+  acl      = "private"
+  region   = "eu-west-2"
+
+  versioning {
+    enabled = true
+  }
+
+  lifecycle_rule {
+    enabled = true
+
+    expiration {
+      days = 7
+    }
+  }
+
+  policy = <<POLICY
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Principal": {"AWS": "arn:aws:iam::${var.aws_backups_account_id}:role/backups"},
+      "Effect": "Allow",
+      "Action": [
+        "s3:ListBucket",
+        "s3:GetBucketLocation"
+      ],
+      "Resource": "arn:aws:s3:::digitalmarketplace-cross-region-database-backups"
+    },
+    {
+      "Principal": {"AWS": "arn:aws:iam::${var.aws_backups_account_id}:role/backups"},
+      "Effect": "Allow",
+      "Action": [
+        "s3:GetObject"
+      ],
+      "Resource": "arn:aws:s3:::digitalmarketplace-cross-region-database-backups/*"
+    }
+  ]
+}
+POLICY
+}
+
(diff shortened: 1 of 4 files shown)


terraform/accounts/backups/s3_buckets.tf
```diff
@@ -1,3 +1,113 @@
+# IAM policy for all backups buckets. Backups role should have list bucket, get bucket and get object.
+data "aws_iam_policy_document" "cross_region_database_backups_s3_bucket_policy_document" {
+  statement {
```
Wynndow: Isn't this duplication of what you've defined in `access.tf`?

terraform/accounts/backups/s3_buckets.tf
```diff
@@ -1,3 +1,113 @@
...
+      "arn:aws:s3:::digitalmarketplace-database-backups/*",
+    ]
+  }
+}
+
+# Extend cross-region policy (using source_json) with Jenkins access for regular backups bucket
+# Jenkins needs list bucket, get bucket, get object and put object.
+# See https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html#example-with-source-and-override
+data "aws_iam_policy_document" "database_backups_s3_bucket_policy_document" {
+  source_json = "${data.aws_iam_policy_document.cross_region_database_backups_s3_bucket_policy_document.json}"
```
Wynndow: Does this extend the policy defined above? Nifty...

benvand:
@Wynndow Had to switch back to inline json policies because I couldn't get that other override thing to work :( Never mind. TBF it's slightly fewer lines 🤷‍♂️

PR #433, opened 2018-09-06
Stop S3 post URL data being echoed in Jenkins logs

lfdebrux:
This PR does a couple of things to try and address the ticket [Jenkins credentials shown in plain text in job scripts and logs](https://trello.com/c/eBl3yUUT). It is presented as an alternative to #427, although I'm not sure it is exactly what @katstevens had in mind when she suggested using a python script...

At its core, this PR adds a feature to the `generate-paas-manifest` script that allows variables to be added to a template from the environment. In `docker run` style, if you want a template to use an environment variable, it has to be specified in the command line arguments as `-v ENV_VAR`.

This feature is then used to get the s3 post body from the environment when the manifest is initially generated, rather than adding it to the apps envvars later.

I also did some general cleaning up of the manifest itself, moving some command line flags from `cf push` into the manifest file, as this seemed less noisy to me.


scripts/generate-paas-manifest.py
```diff
@@ -19,7 +19,9 @@
               help="Output file, if empty the template content is printed to the stdout")
 @click.option('--vars-file', '-f', multiple=True, type=click.Path(exists=True),
               help="Load YAML or JSON variable file")
-def paas_manifest(environment, app, vars_file, out_file):
+@click.option('--var', '-v', multiple=True, type=str,
```
risicle: `-v` to me is `--verbose` but hey whatever

lfdebrux: I'm happy to change it to something else

scripts/generate-paas-manifest.py
```diff
@@ -31,6 +33,20 @@ def paas_manifest(environment, app, vars_file, out_file):
 
     variables = merge_dicts(variables, variables[app])
 
+    # read variables from command line
+    # these can be key-value pairs, or
+    # the name of an environment variable
```
risicle: Maybe this could be mentioned in the actual `help` string?

lfdebrux: Done

dmaws/utils.py
```diff
@@ -103,6 +103,7 @@ def __getitem__(self, key):
 def template_string(string, variables, templates_path=None):
     jinja_env = jinja2.Environment(
         trim_blocks=True,
+        lstrip_blocks=True,
```
katstevens: It might be an idea to update Jinja to 2.10 to bring in line with the FE apps (the AWS repo has it at 2.8 currently) to avoid running into further 'why is Jinja behaving weirdly' gotchas in future...

scripts/generate-paas-manifest.py
katstevens: To be honest I think I'd prefer grabbing each var explicitly so we know what's expected/allowed. 

Not for this PR but: why are we using `click.option` for this script? We use docopt in our other scripts, we might as well be consistent and use it here too? Or indeed use `click.option` everywhere (docopt is not exactly my favourite tool!).

dmaws/utils.py
lfdebrux: Yeah definitely agree...

scripts/generate-paas-manifest.py
lfdebrux: I'm not sure what you mean by explicitly here?

lfdebrux: `click` is nice but I don't think it's doing much heavy lifting here, looking at it I think we could probably implement the same thing using `argparse` and save ourselves a dependency.

katstevens: I mean having separate options for `--dump-file-name`, `--s3-post-url-data` etc. Long winded and less flexible, but also less risky that the wrong thing will end up in the manifest.

lfdebrux: Oh that makes sense, I totally get where you're coming from now.

My worry with doing it this way is then we have to specify the arguments in two places; what about instead we looked at raising errors if the manifest template doesn't use one or more of the cars?

katstevens: I think we'd run in to trouble with the other Makefile commands that use this function. Unless we had a separate `s3_app_paas_manifest` function that checked the vars, then called this one? (Ugh!)

The reason I'm being picky about this is that our usage of paas manifests (for which apps, what is expected) is not well documented or tested at the moment (I have that ticket to make example manifests of course!). Perhaps we should just have a comment here describing what's going on and the sort of vars we'd expect.

lfdebrux:
I've added some user-friendly error-handling to the `generate-paas-manifest` script, to try and catch the most common mistakes.

## Example output

If the user specifies a variable as a command line flag without a value, but that variable isn't set in the environment

```
 (venv) gds5506:digitalmarketplace-aws laurencedebruxelles$ ./scripts/generate-paas-manifest.py -v DUMP_FILE_NAME preview db-backup
Error: Command line variable "--var 'DUMP_FILE_NAME'" was not set by the flag and was not found in environment variables. Please check your environment is correctly configured.
```

---


If the template requires a variable but it isn't found in any of the locations given on the command line:

```
 (venv) gds5506:digitalmarketplace-aws laurencedebruxelles$ ./scripts/generate-paas-manifest.py preview db-backup
Error: The template 'paas/db-backup.j2' thinks that the variable 'DUMP_FILE_NAME' is undefined. Please check you have included all of the var files and command line vars that you need.
```

PR #438, opened 2018-09-12
router paas env vars: set `DM_ANTIVIRUS_API_URL`

risicle:
A prerequisite for https://github.com/alphagov/digitalmarketplace-router/pull/28

### DIFF
--- a/paas/router.j2
+++ b/paas/router.j2
@@ -10,6 +10,7 @@
 
       DM_API_URL: 'https://dm-api-{{ environment }}.cloudapps.digital'
       DM_SEARCH_API_URL: 'https://dm-search-api-{{ environment }}.cloudapps.digital'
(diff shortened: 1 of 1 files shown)


paas/router.j2
```diff
@@ -10,6 +10,7 @@
 
       DM_API_URL: 'https://dm-api-{{ environment }}.cloudapps.digital'
       DM_SEARCH_API_URL: 'https://dm-search-api-{{ environment }}.cloudapps.digital'
+      DM_ANTIVIRUS_API_URL: 'https://dm-antivirus-api-{{ environment }}.cloudapps.digital'
```
katstevens: You'll probably want to add this var to  the `Makefile` nginx linting command as well (and maybe `docker-compose.yml.example`?)

katstevens: Wait those are in the router repo. I've got myself confused 😸

PR #440, opened 2018-09-13
Create S3 buckets for admin reports

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

Three shiny new buckets:
`digitalmarketplace-reports-preview-preview`
`digitalmarketplace-reports-staging-staging`
`digitalmarketplace-reports-production-production`

Copies permissions/policy from the Agreements buckets (which has no explicit policy for the staging bucket).

### DIFF
--- a/terraform/environments/production/variables.tf
+++ b/terraform/environments/production/variables.tf
@@ -19,6 +19,7 @@ variable "g7_draft_documents_s3_url" {}
 variable "documents_s3_url" {}
 variable "agreements_s3_url" {}
 variable "communications_s3_url" {}
+variable "reports_s3_url" {}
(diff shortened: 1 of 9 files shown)


terraform/environments/production/variables.tf
```diff
@@ -19,6 +19,7 @@ variable "g7_draft_documents_s3_url" {}
 variable "documents_s3_url" {}
 variable "agreements_s3_url" {}
 variable "communications_s3_url" {}
+variable "reports_s3_url" {}
```
katstevens: We need to add this to the router template I think.

PR #442, opened 2018-09-17
148-remove-extraneous-variables-in-environments-variablestf-files

benvand:
## Remove unused variables

It would appear that these have been at least partially copied from account/main/vairiables.tf?

We don't need that many variables injected in to these terraform modules so this commit strips un-needed variables
to reduce confusion and prevent extraneous code.

https://trello.com/c/vSknqk9R/148-remove-extraneous-variables-in-environments-variablestf-files

### DIFF
--- a/terraform/environments/preview/variables.tf
+++ b/terraform/environments/preview/variables.tf
@@ -1,34 +1,8 @@
 variable "aws_main_account_id" {}
 variable "aws_dev_account_id" {}
 
-variable "admin_user_ips" {
-  type = "list"
-}
-
-variable "dev_user_ips" {
-  type = "list"
-}
-
-variable "user_ips" {
-  type = "list"
-}
-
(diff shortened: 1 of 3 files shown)


benvand:
screen shot 2018-09-17 at 09 35 43

terraform/environments/preview/variables.tf
```diff
@@ -1,34 +1,8 @@
...
-}
-
-variable "ssh_key_name" {}
-
-variable "g7_draft_documents_s3_url" {}
-variable "documents_s3_url" {}
-variable "agreements_s3_url" {}
-variable "communications_s3_url" {}
-variable "reports_s3_url" {}
-variable "submissions_s3_url" {}
```
katstevens: Aren't these used in the router nginx template? That's why I added `reports_s3_url` last week...

PR #443, opened 2018-09-17
Jenkins IAM user can read/write preview/staging report buckets

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

Enables the Jenkins role to access the preview and staging buckets, so we can test the script (and maybe even run functional tests in future).

I've already run terraform plan/apply for this so the changes reflect reality.

### DIFF
--- a/terraform/environments/preview/s3_buckets.tf
+++ b/terraform/environments/preview/s3_buckets.tf
@@ -60,7 +60,10 @@ data "aws_iam_policy_document" "reports_bucket_policy_document" {
     effect = "Allow"
 
     principals {
-      identifiers = ["arn:aws:iam::${var.aws_dev_account_id}:role/developers"]
+      identifiers = [
+        "arn:aws:iam::${var.aws_dev_account_id}:role/developers",
+        "arn:aws:iam::${var.aws_main_account_id}:role/jenkins-ci-IAMRole-1FIPDG9DE2CWJ",
+      ]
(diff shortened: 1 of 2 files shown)


terraform/environments/preview/s3_buckets.tf
```diff
@@ -60,8 +60,12 @@ data "aws_iam_policy_document" "reports_bucket_policy_document" {
     effect = "Allow"
 
     principals {
-      identifiers = ["arn:aws:iam::${var.aws_dev_account_id}:role/developers"]
-      type        = "AWS"
+      identifiers = [
+        "arn:aws:iam::${var.aws_dev_account_id}:role/developers",
+        "arn:aws:iam::${var.aws_main_account_id}:role/jenkins-ci-IAMRole-1FIPDG9DE2CWJ",
```
lfdebrux: Another annoying question; why are two roles needed?

katstevens: The first one is the dev group, i.e. us - so I can access/debug the bucket in the console.
The second one is the Jenkins role - the script needs access to upload the file. We can probably remove this once the job is fully tested/run in prod.

lfdebrux: Right, thanks

PR #445, opened 2018-09-17
Stream antivirus sns feedback to Kibana

Wynndow:
For [this trello ticket](https://trello.com/c/CXxL1M8x).

SNS creates logs in CloudWatch to give us feedback on the sucess or
failure of its notifications. This is useful stuff and we should make it
more accessible by streaming it out to Kibana.


risicle:
Is there anything that can be added to the manual as a result of this?

PR #456, opened 2018-09-25
Remove temporary users from terraform

benvand:


katstevens:
To check my understanding - the non-temp users are described in `modules/paas` (rather than say, `modules/iam-users`)?

benvand:
Yeah so the plan will read something like `- aws_iam_user.paas_app` rather than `module.paas.aws_iam_user.paas_app`

PR #457, opened 2018-09-26
171-remove-packer-roles-from-terraform-no-longer-used

benvand:
As per commit https://github.com/alphagov/digitalmarketplace-aws/commit/83f28b00a11bfbf7c6ec65064f8472c02f5b8c6a

> Remove packer templates and Makefile
> We're no longer using packer to deploy AMIs.
>  master (#348)
> @allait
> allait committed on 24 Oct 2017

Use of packer for deployment was removed, we should remove the roles from terraofrm.
Checked here:
https://console.aws.amazon.com/iam/home?region=eu-west-1#/roles/packer?section=access_advisor

Last used in main account 344 days ago
Last used in development account Never
Last used in production account Never

https://trello.com/c/ANvNGWOj/171-remove-packer-roles-from-terraform-no-longer-used


terraform/modules/iam-users/developers.tf
```diff
@@ -18,18 +18,6 @@ resource "aws_iam_group_policy" "developers" {
...
-      "Effect": "Allow",
-      "Action": [
-        "sts:AssumeRole"
-      ],
-      "Resource": "arn:aws:iam::*:role/packer",
-      "Condition": {
-        "Bool": {
-          "aws:MultiFactorAuthPresent": true
-        }
-      }
```
lfdebrux: A small pet peeve, we can leave trailing commas for the last item in a list in terraform, so that diffs are cleaner.

It is just a personal taste thing though so please feel free to ignore me.

PR #458, opened 2018-09-27
Cross region replication for document buckets

katstevens:
Trello: https://trello.com/c/TaKJj9U3/141-cross-region-replicate-document-buckets

Now ready for final review:
- created new replication roles/policies for the development and production accounts
- created a new `eu-west-2` provider for Preview and Production
- created new destination S3 buckets for Preview and Production
- added the replication config to the source buckets for Preview and Production

### DIFF
--- a/terraform/environments/preview/s3_buckets.tf
+++ b/terraform/environments/preview/s3_buckets.tf
@@ -51,6 +51,20 @@ resource "aws_s3_bucket" "agreements_bucket" {
   }
 
   policy = "${data.aws_iam_policy_document.agreements_bucket_policy_document.json}"
+
+  replication_configuration {
+    role = "${aws_iam_role.replication_role.arn}"
+
+    rules {
+      id     = "cross-region-agreements-replication"
+      prefix = ""
+      status = "Enabled"
+
+      destination {
+        bucket = "${aws_s3_bucket.cross_region_agreements_s3_bucket.arn}"
+      }
+    }
+  }
 }
 
 # Reports - devs: read write list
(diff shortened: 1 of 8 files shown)


terraform/environments/preview/s3_buckets.tf
```diff
@@ -212,3 +212,178 @@ resource "aws_s3_bucket" "submissions_bucket" {
...
+        "s3:ListBucket",
+        "s3:GetBucketLocation"
+      ],
+      "Resource": "arn:aws:s3:::digitalmarketplace-cross-region-documents-preview-preview"
+    },
+  ]
+}
+POLICY
+
+  replication_configuration {
```
benvand: The replication configuration needs to live on the bucket you're replicating from, the source, rather than the destination.

In the backups terraform I defined the destination bucket above the source bucket because semantically it makes sense to have the  replication bucket exist before you can tell something to replicate in to it.

terraform/environments/preview/s3_buckets.tf
```diff
@@ -212,3 +212,178 @@ resource "aws_s3_bucket" "submissions_bucket" {
...
+  }
+
+  }
+
+  policy = <<POLICY
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Principal": {"AWS": "arn:aws:iam::${var.aws_development_account_id}:role/replication"},
```
benvand: The role `replication` gets its permissions from 2 places in the replication set up. The first is in the `replication_policy` attached to the role (looks good btw 👍 ) and the second is by being listed on the source bucket as the role to perform replication. We don't actually need it to have any permissions on the destination bucket other than those supplied by the role policy IIRC if it exists within the same account.

terraform/environments/preview/s3_buckets.tf
katstevens: I've pushed a couple of commits which:
- move the `replication_configuration` blocks to the source buckets
- get rid of the destination bucket policies
- fixes trailing commas on the IAM role lists

I ran `make plan` for the roles in dev/prod and all seems fine; I can't run the plan for the buckets until those roles are created for real, as they're on different Makefiles.

katstevens:
😿 The bucket names need to be under 64 characters! Going to rename the prod ones.

PR #470, opened 2018-10-30
terraform: add "*-dropped-antivirus-sns" log metrics, sourced from both failure and success sns feedback logs

risicle:
https://trello.com/c/cQEkAi5b/214-enable-hosted-graphite-alerts-for-final-failed-sns-deliveries

Triggered by both "final retry" "failure"s and 4xx responses from the "success" log group.

This includes quite a bit of variable wiring around the place so we can pull e.g. the max number of retries to filter for and the names of the log groups from the various places they're decided.

### DIFF
--- a/terraform/environments/preview/main.tf
+++ b/terraform/environments/preview/main.tf
@@ -52,10 +52,13 @@ module "log_streaming" {
 }
 
 module "log_metrics" {
-  source                = "../../modules/log-metrics"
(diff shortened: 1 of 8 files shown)


risicle:
(bloody terraformat...)

risicle:
Well yesterday @Wynndow said it looked "lovely", so there's your second opinion.

PR #471, opened 2018-10-31
hosted graphite alerts: add "Production antivirus 'dropped' SNS messages" alert

risicle:
https://trello.com/c/cQEkAi5b/214-enable-hosted-graphite-alerts-for-final-failed-sns-deliveries

It even works.


risicle:
Stylistic, but there is a real reason why I prefer them I'll go into another time...

PR #472, opened 2018-11-08
Example PaaS manifests for API, Search API, Buyer FE and Router

katstevens:
Trello: https://trello.com/c/GNMyCgSJ/232-example-paas-manifest-yaml-files

Generated some example PaaS manifests for Preview (with all credentials etc stripped) for:
- API
- Search API
- Router
- Buyer FE (the other frontend apps are similar enough that I only did this one - happy to add more if people think there's value in it)

These manifests are unlikely to change drastically in the future, but if they do, I've included a line at the top of the examples on how to generate new ones.

I've left in values for routes and URLs as these are public (described in the `vars` folder of this repo). Please double check that I've taken everything else out!


katstevens:
Ta! 👍

PR #474, opened 2018-11-30
Enable access logging on Jenkins load balancers

lfdebrux:
We want to be able to audit who is accessing our Jenkins build server, and when.

This PR enables access logging on the Elastic Load Balancer that sits in front of our Jenkins box; this means that all SSH and HTTPS access will be logged.

The logs are stored hourly in an S3 bucket, in a compressed format.

For more details on how this works, see the [Trello ticket][ticket].

The Terraform for these changes have already been tested and applied, and the logging is now taking place.

[ticket]: https://trello.com/c/S7iFoAU8/253-configure-jenkins-box-ssh-access-logging

### DIFF
--- a/terraform/modules/jenkins/s3_bucket.tf
+++ b/terraform/modules/jenkins/s3_bucket.tf
@@ -1,0 +1,28 @@
+resource "aws_s3_bucket" "jenkins_logs_bucket" {
+  bucket        = "${var.name}-${var.dns_name}-logs-bucket"
+  force_destroy = true
+
+  policy = <<POLICY
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "AWSElasticLoadBalancerWrite",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    "arn:aws:iam::156460612806:root"
+                ]
+            },
+            "Action": "s3:PutObject",
+            "Resource": "arn:aws:s3:::${var.name}-${var.dns_name}-logs-bucket/*",
+            "Condition": {
+                "StringEquals": {
+                    "s3:x-amz-acl": "bucket-owner-full-control"
+                }
+            }
+        }
+    ]
+}
+POLICY
+}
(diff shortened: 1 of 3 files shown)


terraform/modules/jenkins/s3_bucket.tf
```diff
@@ -0,0 +1,28 @@
...
+  policy = <<POLICY
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "AWSElasticLoadBalancerWrite",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    "arn:aws:iam::156460612806:root"
```
katstevens: Which user is this? Should we be hardcoding this ID?

terraform/modules/jenkins/s3_bucket.tf
```diff
@@ -0,0 +1,28 @@
+resource "aws_s3_bucket" "jenkins_logs_bucket" {
+  bucket        = "${var.name}-${var.dns_name}-logs-bucket"
+  force_destroy = true
+
+  policy = <<POLICY
+{
+    "Version": "2012-10-17",
```
katstevens: We have `Version: 2012-10-17` all over our terraform stuff and I am not sure what it refers to. Terraform version? The policy version? AWS resource version?

terraform/modules/jenkins/s3_bucket.tf
lfdebrux: Ah, good spot.

This is an annoying thing about the Elastic Load Balancer, the way to include it in a policy is just to include this account ID from [this table](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-access-logs.html#w281aac21b9c17c11b9b8b3b5b3) in the AWS docs.

It would be nice to document this somewhere, but I don't think you can put comments in JSON...

terraform/modules/jenkins/s3_bucket.tf
lfdebrux: I know right!

I think it's policy document specification version. It's in all the AWS documentation examples. We could try leaving it out...

lfdebrux: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

lfdebrux: Reading The Fine Documentation reveals that if we don't specify a version, we lose out on newer features... 🤷‍♀️

terraform/modules/jenkins/s3_bucket.tf
lfdebrux: Comment added to just above the policy definition

PR #478, opened 2018-12-14
289 replace hg alerts with cloudwatch alerts

benvand:
I've moved the `log-metric-filters` module in to a `logging` directory where the 2nd commit then introduces the alarms modules. So the directory structure looks like:

screen shot 2018-12-14 at 14 25 20

Then added an alarms sub-directory where we can add specific alarms.

I've recreated what we have for _production_ on HG for _preview_ on CloudWatch


terraform/modules/logging/alarms/email-notification/sns.tf
```diff
@@ -0,0 +1,107 @@
...
+}
+
+resource "aws_sns_topic_policy" "alarm_email_topic_policy" {
+  arn = "${aws_sns_topic.alarm_email_topic.arn}"
+
+  policy = "${data.aws_iam_policy_document.alarm_email_topic_policy_document.json}"
+}
+
+// The below is unsupported but describes the subscription of a slack email to the topic
+// https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html#email
```
risicle: ...and so we should use the aws console to do this if we ever need to replace it?

risicle: I would say leave it in.

PR #488, opened 2019-01-11
Remove Hosted Graphite configuration

lfdebrux:
We're no longer using Hosted Graphite, so I think these files are no longer needed.

We might want to keep around some code for pushing Grafana dashboards for our new metrics system, but my understanding was that we can't use our existing solution?

Thoughts welcome.


katstevens:
I have a question - where are we defining the new Grafana dashboards?

lfdebrux:
> I have a question - where are we defining the config for the new Grafana dashboards?

Currently it's not stored anywhere. We can manually export/import each dashboard by copy-pasting the JSON text https://grafana-paas.cloudapps.digital/d/B3XEtGPik/dm-overview?editview=dashboard_json&refresh=5m&orgId=1&panelId=&fullscreen=&edit=

katstevens:
OK... Can we replace the `grafana/dm-apps.json` we're deleting here with that config, even if it's just temporary and we have to manually c&p it in, just so that we have it in version control somewhere? We can probably call it something else to make it clear it's not the old one.

lfdebrux:
> OK... Can we replace the `grafana/dm-apps.json` we're deleting here with that config, even if it's just temporary and we have to manually c&p it in, just so that we have it in version control somewhere? We can probably call it something else to make it clear it's not the old one.

I've added these in a separate commit.

PR #491, opened 2019-01-16
Deployment requirements for metrics

benvand:
* Add explicitly defined route (/metrics) for metrics
  * This route allows us to bind the whitelisting route service
* Bind digitalmarketplace_prometheus service using the app manifest


benvand:
At the moment yeah

PR #496, opened 2019-02-06
clean-db-dump.sql: perform construction of mock json in a safer way than pure string formatting

risicle:
https://trello.com/c/JgsizEmz

While I don't _think_ the exact fields being used here _should_ cause us to be at risk, all it takes is for someone less experienced to need to include an extra field's data, and decide to simply copy & paste an existing line and include a text-based field - at this point we would open ourselves to the string being able to escape from the confines of its intended json key using the `"` character.

I've run this against an _already cleaned_ version of the database on my local machine, because that's all I think I'm able to do. It appears to do the right thing. Anyone got any bright ideas as to how to safely test this against the real target short of merging it & trying it for reals?

### DIFF
--- a/scripts/clean-db-dump.sql
+++ b/scripts/clean-db-dump.sql
@@ -109,14 +109,18 @@ FROM audit_events;
 -- Sanitise framework agreements because they contain personal data
 \echo 'Update framework agreements'
 UPDATE framework_agreements
-SET signed_agreement_details = ('{"signerName": "A. Nonymous",
-                                  "signerRole": "The Boss",
-                                  "uploaderUserId": ' || TEXT(signed_agreement_details->'uploaderUserId') || ',
-                                  "frameworkAgreementVersion": ' || TEXT(signed_agreement_details->'frameworkAgreementVersion') || '}'
-                               )::json,
+SET
+    signed_agreement_details = json_build_object(
+        'signerName', 'A. Nonymous',
+        'signerRole', 'The Boss',
+        'uploaderUserId', signed_agreement_details->'uploaderUserId',
+        'frameworkAgreementVersion', signed_agreement_details->'frameworkAgreementVersion'
+    ),
     signed_agreement_path = 'not/the/real/path.pdf'
-WHERE signed_agreement_details IS NOT NULL
-AND cast(signed_agreement_details AS TEXT) != 'null';
+WHERE
+        signed_agreement_details IS NOT NULL
+    AND
+        json_typeof(signed_agreement_details) != 'null';
 
 UPDATE framework_agreements
 SET countersigned_agreement_path = 'not/the/real/path.pdf'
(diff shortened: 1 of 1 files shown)


scripts/clean-db-dump.sql
```diff
@@ -150,31 +155,25 @@ INSERT INTO brief_responses (
...
-            '], '
-        END ||
-        '"availability": "09/09/17", ' ||
-        '"essentialRequirementsMet": true, ' ||
-        '"respondToEmailAddress": "example-email@example.gov.uk"' ||
-      '}'
-    )::JSON,
+    -- create 'data' column with varying numbers of evidence
+    json_strip_nulls(json_build_object(
+        'essentialRequirements', array_fill('{"evidence": "Some essential evidence."}'::json, ARRAY[essential_requirements_len]),
```
lfdebrux: Is it possible to use json_build_object here? Obviously not necessary but I'm curious.

lfdebrux:
This seems like a good idea, I'm just trying to get my head around the syntax

scripts/clean-db-dump.sql
risicle: It is using `json_build_object`, it's just wrapped in `json_strip_nulls` too so I can choose to "omit" keys by simply outputting them as `null`. It's less painful than other syntaxes to do this.

risicle:
Do we think I should just try merging it? If it's wrong, the first thing to break will be the functional tests.

risicle:
Also I don't know if people noticed, but one of the changes I made was from

```
MOD(eligible_brief_supplier_pairings.supplier_id, 5) = 0 AND MOD(eligible_brief_supplier_pairings.supplier_id, 3) = 0
```
which I looked at and thought ... well, that's just

```
MOD(eligible_brief_supplier_pairings.supplier_id, 30) = 0
```

isn't it? Was I missing something?

scripts/clean-db-dump.sql
```diff
@@ -126,18 +130,19 @@ WHERE countersigned_agreement_path IS NOT NULL;
...
+                'primaryContactEmail', 'supplier-user@example.com'
+            )
+    END
 WHERE declaration IS NOT NULL
-  AND declaration::varchar != 'null'
-  AND declaration::varchar != '{}';
+  AND json_typeof(declaration) != 'null'
+  AND declaration::text != '{}';
 
 -- Create some fake draft and submitted brief responses on closed briefs using eligible suppliers
```
katstevens: Running this locally I noticed this fake brief responses bit doesn't have an `echo` line for the output, hopefully it's not too out of scope to add one in 😸

benvand:
Ah the MOD does work differently, will write up now

benvand:
So the reason I wrote it separately (it can be written as `x % (3*5) == 0`) is because below we set the `submitted_at` to none based on %5.

So 1 in 5 brief responses created will be drafts.

Back above we _blank_ the nice to have requirements based on
1. if the brief has no nice to haves
2. the response is in 1 in 3rd of the '1 in 5 drafts'

This way only drafts will have nice to haves not filled in, and only 1/3rd of those drafts.

| supplier id  | is submitted | with nice to haves |
| ------------ | ------------ | ------------------ |
| 1            | y            | y                  |
| 2            | y            | y                  |
| 3            | y            | y                  |
| 4            | y            | y                  |
| 5            | n            | y                  |
| 6            | y            | y                  |
| 7            | y            | y                  |
| 8            | y            | y                  |
| 9            | y            | y                  |
| 10           | n            | y                  |
| 11           | y            | y                  |
| 12           | y            | y                  |
| 13           | y            | y                  |
| 14           | y            | y                  |
| 15           | n            | n                  |

benvand:
All the other bits look cool though 👍

risicle:
Yeah, so I saw that and considered whether it was just to make it more obvious that the `% 5` coincided with the `submitted_at` - it's probably better to leave it as it was so people don't have to do the arithmetic themselves to notice, but i'll leave a correction to the comment. And yeah, duh, the LCM of 3 and 5 is 15, not 30.

PR #497, opened 2019-02-08
clean-db-dump.sql: populate more fields of declaration with pseudorandom data

risicle:
https://trello.com/c/HFygifh6

Using a smattering of declaration fields from frameworks **past & present**, taking care to only add fields which were actually present on an original declaration, but not based on any actual answers in any way. The pseudorandomness is seeded on the `supplier_id`, `framework_id` combo, so should be stable. Using `md5(...` as pseudorandom function as it's the most readily available tool in postgres.

I'd be interested if anyone can think of any more fields they think should be included.

### DIFF
--- a/scripts/clean-db-dump.sql
+++ b/scripts/clean-db-dump.sql
@@ -126,7 +126,7 @@ UPDATE framework_agreements
 SET countersigned_agreement_path = 'not/the/real/path.pdf'
 WHERE countersigned_agreement_path IS NOT NULL;
 
--- Overwrite declarations with the smallest possible valid entry
+-- Overwrite declarations with a small but hopefully valid entry
 -- Removes all personal data while keeping our app working as expected
 \echo 'Blank out declarations'
 UPDATE supplier_frameworks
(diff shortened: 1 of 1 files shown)


scripts/clean-db-dump.sql
```diff
@@ -135,12 +135,61 @@ SET
...
+        supplier_id,
+        framework_id,
+        md5(supplier_id::text || '|' || framework_id::text) AS id_hash,
+        upper(translate(md5(supplier_id::text || '|' || framework_id::text), '0123456789', 'IOUYGHJKLM')) AS id_hash_ualpha,
+        translate(md5(supplier_id::text || '|' || framework_id::text), 'abcdef', '987654') AS id_hash_digits
+    FROM supplier_frameworks
+) AS annotation
+WHERE supplier_frameworks.supplier_id = annotation.supplier_id AND supplier_frameworks.framework_id = annotation.framework_id
+  AND declaration IS NOT NULL
   AND json_typeof(declaration) != 'null'
```
benvand: Could move these in to the sub query for efficiency but doesn't make much odds at all 👍

katstevens:
Can we include some fake address fields? The script we tidied last week used the templates in the agreements repo, which has stuff like `supplierRegisteredBuilding`, `supplierRegisteredTown`, `supplierRegisteredPostcode` etc.

scripts/clean-db-dump.sql
risicle: I _think_ the query planner's able to perform that optimization itself

risicle:
Good call @katstevens I was just holding off them because I was expecting someone to tell me I've gone too far if I did that...

lfdebrux:
I just realised that address details would also be useful for the admin feature @katstevens and I are working on right now

risicle:
I assume you mean `registeredAddressBuilding` et al - was having trouble finding `supplierRegisteredBuilding` in the frameworks repo and we appear to have removed these questions entirely from G10 on. Or do you mean the actual supplier contact details that live on the `contact_information` table? Because those are already filled out with some dummy values.

katstevens:
Hmm, those fields are definitely in the G10 declarations. We don't have address fields at supplier level so we must ask for it during applications somewhere.

Here's the template where they're used (to generate the agreement PDFs) https://github.com/alphagov/digitalmarketplace-agreements/blob/master/documents/g-cloud-10/framework-agreement-signature-page.html

lfdebrux:
I think the script is using some sort of backwards compatibility shim?

https://github.com/alphagov/digitalmarketplace-api/blob/master/app/supplier_constants.py

risicle:
That's a 4-repo reference chain you had to chase just there - classic digital marketplace.

So basically those variables are injected into the context of the template _from_ the supplier object.

lfdebrux:
> So basically those variables are injected into the context of the template _from_ the supplier object.

The way it actually gets into the template is super complicated, but I'm fairly sure it uses declaration address details, not supplier contact information address details.

katstevens:
Yep - the supplier can change their contact details later (eg for a subsequent FW application) so we copy a snapshot into the declaration at the point in the application process where they confirm their company details. The agreement PDF wants that snapshot.

risicle:
Well seeing as `supplier*` is actually stored in the declaration I'll add it too... `registeredAddressBuilding` et al are only for old frameworks.

PR #504, opened 2019-03-04
377-aws-alarm-for-disk-space-on-jenkins

benvand:
* Add permissions for jenkins instance to PutMetricData in to Cloudwatch (used with cloudwatch-logs-agent)
* Add new alarms for disk space on jenkins

https://trello.com/c/PdXsLT6d/377-aws-alarm-for-disk-space-on-jenkins

### DIFF
--- a/terraform/accounts/main/alarms.tf
+++ b/terraform/accounts/main/alarms.tf
@@ -1,0 +1,63 @@
+module "alarm_email_sns" {
+  source     = "../../modules/logging/alarms/email-notification"
+  name       = "main"
+  account_id = "${var.aws_main_account_id}"
+}
+
+module "alarm_recovery_email_sns" {
+  source     = "../../modules/logging/alarms/email-notification"
+  name       = "main-recovery"
+  account_id = "${var.aws_main_account_id}"
+}
+
+resource "aws_cloudwatch_metric_alarm" "jenkins_data_volume_disk_space" {
+  alarm_name        = "jenkins-data-volume-disk-space"
+  alarm_description = "Alerts on disk space on 100gb /data volume"
+
+  // Metric
+  namespace   = "CWAgent"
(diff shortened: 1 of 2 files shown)


terraform/accounts/main/alarms.tf
```diff
@@ -0,0 +1,63 @@
...
+  comparison_operator = "LessThanOrEqualToThreshold"
+  threshold           = "10000000000"
+
+  // Email slack
+  alarm_actions = ["${module.alarm_email_sns.email_topic_arn}"]
+  ok_actions    = ["${module.alarm_recovery_email_sns.email_topic_arn}"]
+}
+
+resource "aws_cloudwatch_metric_alarm" "jenkins_main_volume_disk_space" {
+  alarm_name        = "jenkins-data-volume-disk-space"
```
katstevens: Should this be a different name? `jenkins-main-volume-disk-space`?

terraform/accounts/main/jenkins_instance_profile.tf
```diff
@@ -159,9 +159,28 @@ data "aws_iam_policy_document" "cloudwatch-policy" {
...
+}
+
+data "aws_iam_policy_document" "cloudwatch-metrics-policy" {
+  statement {
+    actions = [
+      "cloudwatch:PutMetricData",
+    ]
+
+    resources = [
+      "*",
```
katstevens: Does this mean 'everything'? Or are we waiting on a hardcoded value?

benvand: Means everything, and according to the AWS interface has to be everything.

terraform/accounts/main/alarms.tf
benvand: Nice catch 👍 updated

PR #505, opened 2019-03-04
deploy-app: add network policy handling, enable for preview

risicle:
https://trello.com/c/H53K45It

This does a few things, but luckily should only have an effect on preview for now:

 - adds secondary `.apps.internal` routes to each of the preview api apps
 - adds a new step in the `make deploy-app` process which will attempt to create the network policies the newly created app is going to need. Do read the extensive comments on this, because it's quite carefully considered (!) and should hopefully work even in the face of simultaneous release procedures of both parties of an ingress/egress relationship.
 - these ingress/egress relationships are defined in the vars yaml on the "egress" side of the relationship by defining a `egress_to_applications` key. These should automatically get picked up as ingresses for the _referred_ app using a little bit of `yq` magick. Again, here i've only defined these egress relationships for the preview stage and designed the relevant steps so that they're no-ops for stages that have nothing relevant defined.

Note this PR doesn't actually make any apps actually _use_ the private urls yet - that comes later once we're sure bringing up the routes and policies works reliably enough. The idea is I merge this and then that allows me to hit "re-release all apps" for preview, which once I've done will allow us to rely on this new setup being persistent (when releasing from my machine without the code upstreamed, any random releases resulting from app merges will knock my alterations out).

I've also added a `add-all-app-network-policies` for panic-mode repairing the network policies if it turns out I haven't got the release procedure right.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,15 @@ deploy-app: ## Deploys the app to PaaS
 	$(if ${STAGE},,$(error Must specify STAGE))
 	cf push --no-start -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME}
 
+	# apps that communicate with each other over the "internal" private network (*.apps.internal) need
+	# to have their explicitly "allowed" communication paths added as "network policies". these must be
+	# added before the app is started otherwise it will not be able to access those other apps, or those
+	# other apps won't be able to access *it*. the policies end up being associated with the app's guid
+	# rather than app name, so these need to be re-created for new apps we create, but the policies
+	# should follow the app's renaming at the end of the release process.
+	#
 	# the order of operations performed here (including inside add-application-network-policies.sh)
-	# is carefully designed to make it impossible for an app to get to the point of being started
+	# is carefully designed to make it hopefully-impossible for an app to get to the point of being started
 	# with any required network policies not being in place, even if the "other" app (the other "party"
 	# in the policy) is simultaneously going through the release process and having its names & routes
 	# similarly juggled around. consider this carefully if making any changes to this.
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,14 @@ paas-login: ## Log in to PaaS
 	$(if ${PAAS_SPACE},,$(error Must specify PAAS_SPACE))
 	@cf login -a "${PAAS_API}" -u ${PAAS_USERNAME} -p "${PAAS_PASSWORD}" -o "${PAAS_ORG}" -s "${PAAS_SPACE}"
 
+.PHONY: add-all-app-network-policies
+add-all-app-network-policies: ## attempts to (re-)add all known PaaS "network policies" suitable for a particular $STAGE
+	$(call check_space)
+	$(if ${STAGE},,$(error Must specify STAGE))
+	for APPLICATION_NAME in $$(yq -rs '.[0] * .[1] | to_entries | .[] | select(.value | has("egress_to_applications"))? | .key' ./vars/common.yml ./vars/${STAGE}.yml) ; do \
+		./scripts/add-application-network-policies.sh ./vars/$${STAGE}.yml ./vars/common.yml $${APPLICATION_NAME} ; \
+	done
+
 .PHONY: deploy-app
 deploy-app: ## Deploys the app to PaaS
 	$(call check_space)
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,15 @@ deploy-app: ## Deploys the app to PaaS
 	$(call check_space)
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
 	$(if ${RELEASE_NAME},,$(error Must specify RELEASE_NAME))
+	$(if ${STAGE},,$(error Must specify STAGE))
 	cf push --no-start -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME}
+
+	# the order of operations performed here (including inside add-application-network-policies.sh)
+	# is carefully designed to make it impossible for an app to get to the point of being started
+	# with any required network policies not being in place, even if the "other" app (the other "party"
+	# in the policy) is simultaneously going through the release process and having its names & routes
+	# similarly juggled around. consider this carefully if making any changes to this.
+	./scripts/add-application-network-policies.sh ./vars/${STAGE}.yml ./vars/common.yml ${APPLICATION_NAME} "-release"
 
 	@echo "Waiting to ensure new app's assigned service credentials have taken effect..."
 	sleep 60
--- a/scripts/add-application-network-policies.sh
+++ b/scripts/add-application-network-policies.sh
@@ -1,0 +1,35 @@
+#!/bin/bash -e
+
+if [ "$#" -lt 3 -o "$#" -gt 4 ]; then
+    echo "Usage: ./scripts/add-application-network-policies.sh <vars_yaml> <base_yaml> <application_name> [<application_suffix>]"
+    exit 1
+fi
+
+export VARS_YAML=$1
+export BASE_YAML=$2
+export APPLICATION_NAME=$3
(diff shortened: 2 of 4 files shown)


risicle:
As @katstevens noticed and I forgot to mention in the description - this PR *only* moves the APIs to internal addresses. This is because the frontends all live on the same domain under different paths, relying on PaaS' gorouter to do the actual routing to the correct frontend. This can't be done with internal connections because internal connections aren't actually handled on an http level at all - they're raw tcp connections.

We could possibly move the frontends to private addresses at some point if we actually set up the per-path routing on the ... router, directing requests to different frontend apps itself. But for now, they're protected by the basic auth nonsense, which we'll just have to keep in place for the time being.

scripts/add-application-network-policies.sh
```diff
@@ -0,0 +1,35 @@
...
+# we must be tolerant to any of these commands failing because any of these application names may or may not exist at the exact
+# point in time the command is executed. This is ok, because the *other* release process is also expected to be attempting
+# to create these policies at the same time and this routine is designed so that at least one of them should be able to succeed
+# before an application is started.
+#
+# And of course, because of this the commands also need to be robust against the network policy already existing
+
+for EGRESS_APPLICATION_NAME in $(yq -rs '.[0] * .[1] | .[env.APPLICATION_NAME]?.egress_to_applications?[]?' "$BASE_YAML" "$VARS_YAML") ; do
+    echo "Adding policy for access from ${APPLICATION_NAME}${APPLICATION_SUFFIX} to $EGRESS_APPLICATION_NAME-release..."
+    cf add-network-policy ${APPLICATION_NAME}${APPLICATION_SUFFIX} --destination-app $EGRESS_APPLICATION_NAME-release --protocol tcp --port 8080 || echo "...but that's probably fine."
```
katstevens: If we can assume that `$EGRESS_APPLICATION_NAME-release` is always going to have suffix `-release`, can we assume that for the `${APPLICATION_NAME}${APPLICATION_SUFFIX}` too?

risicle: I made `$APPLICATION_SUFFIX` customizable _really_ so that this script could double up for the `add-all-app-network-policies` command, which will need to operate for the actual target application name - in this case `$APPLICATION_SUFFIX` can be left blank.

risicle: (whereas we always need to consider the possibility of the "other party" app being in the middle of a release, so operate against both possibilities)

katstevens:
So, to check my understanding by working through an example, doing a release of the `api` app would involve:

- Creating the `api-release` app as normal, with `cf push`
- Adding the following network policies, which are doubled up in case we're simultaneously releasing `search-api`:

```
cf add-network-policy api-release --destination-app search-api-release
cf add-network-policy api-release --destination-app search-api

cf add-network-policy search-api-release --destination-app api-release
cf add-network-policy search-api --destination-app api-release
```

- Waiting 60 secs for `api-release` to successfully bind to services
- Unmap the routes from `api` and waiting for any in-progress requests to complete
- Carry on as normal (deleting old `api` app etc)

Questions arising:
- Do we ever need to `cf remove-network-policy`? Would this be at the unmap stage?
- Should we add the policies before or after the initial sleep, i.e. will the add-network-policy fail if the new `.apps.internal` route isn't there yet for whatever reason?
- Do we already have `yq` installed on Jenkins?

scripts/add-application-network-policies.sh
katstevens: Cool - makes sense 👍

risicle:
> Adding the following network policies, which are doubled up in case we're simultaneously releasing search-api

Not just that, it would also add corresponding policies for each of the FE apps that rely on it, _and_ the router.

risicle:
> Do we already have yq installed on Jenkins?

Great question. I will find out.

risicle:
> Do we ever need to cf remove-network-policy? Would this be at the unmap stage?

They will die with the app as it's deleted.

risicle:
> Should we add the policies before or after the initial sleep

I don't know - I think we may have to experiment here, because on the other hand it's equally possible that the policies might take a few seconds to have an effect after creation...

Makefile
```diff
@@ -81,13 +81,29 @@ paas-login: ## Log in to PaaS
...
+
 .PHONY: deploy-app
 deploy-app: ## Deploys the app to PaaS
 	$(call check_space)
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
 	$(if ${RELEASE_NAME},,$(error Must specify RELEASE_NAME))
+	$(if ${STAGE},,$(error Must specify STAGE))
 	cf push --no-start -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME}
 
+	# the order of operations performed here (including inside add-application-network-policies.sh)
```
benvand: This comment doesn't give us much info on the _why_ it's done here. Just that it's important it is done here.

Something like:
# This step must be performed after XYZ but before ABC to ensure DEF

Would help other devs get where they need to be quicker

scripts/add-application-network-policies.sh
```diff
@@ -0,0 +1,35 @@
...
+    echo "Adding policy for access from ${APPLICATION_NAME}${APPLICATION_SUFFIX} to $EGRESS_APPLICATION_NAME..."
+    cf add-network-policy ${APPLICATION_NAME}${APPLICATION_SUFFIX} --destination-app $EGRESS_APPLICATION_NAME --protocol tcp --port 8080 || echo "...but that's probably fine."
+done
+
+for INGRESS_APPLICATION_NAME in $(yq -rs '.[0] * .[1] | to_entries | .[] | select(.value.egress_to_applications? | any(. == env.APPLICATION_NAME))? | .key' "$BASE_YAML" "$VARS_YAML") ; do
+    echo "Adding policy for access from $INGRESS_APPLICATION_NAME-release to ${APPLICATION_NAME}${APPLICATION_SUFFIX}..."
+    cf add-network-policy $INGRESS_APPLICATION_NAME-release --destination-app ${APPLICATION_NAME}${APPLICATION_SUFFIX} --protocol tcp --port 8080 || echo "...but that's probably fine."
+    echo "Adding policy for access from $INGRESS_APPLICATION_NAME to ${APPLICATION_NAME}${APPLICATION_SUFFIX}..."
```
benvand: The `... that's probably fine` is ambiguous for anyone using the output to debug.

Can't we check why it's failed? And notify accordingly?

Or failing that can we check at the end that everything we expect to exists exists and exit 1/ 0 the whole thing based on that?

benvand: Or if we're expecting failures then don't print out anything.

Makefile
risicle: I'll take a look and try, but the exact reason for it will probably have a long long explanation for it - it's not something I've actually thought about in terms of english language, it's just that I found this order of operations impossible to trip up no matter where it gets interrupted...

scripts/add-application-network-policies.sh
risicle: Um yeah, I'm not sure what's the best thing to do here - the output is indeed a bit ambiguous, but really the whole operation is a bit ambiguous. It's something that's kind of a cooperation between two release processes. It slightly relies on having the things that fail on one of them succeed on the other. And it's hard to create definite output as to whether that's ended up happily being the case or not... I'm going to think about this too

Makefile
risicle: See, the way I determined the validity of this scheme was really by playing a sort of card game in my text editor trying to find different interleaving orders that would end up starting an app without its policies in place:
screen1

risicle: In fact, the following is a particularly interesting case, because from B's POV, both of its attempted policy additions would have appeared to fail, but actually the end result would end up being fine because _both_ of A's additions of them should have succeeded.
screen1

I _do_ think though I can make the failure messages a bit more illustrative of this.

risicle:
I've improved the "error" messages, take a peek.

risicle:
I'm going to merge this so I can at least get an idea about whether it works as I expect it to when used in anger.

PR #507, opened 2019-03-05
Create new bucket to store dev cleaned db dumps and give access

benvand:
As part of:
https://trello.com/c/uUxYtjSt/345-gdrive-client-alternatives-consolidate-data-dumps-store-in-s3

We need a new bucket that devs can access to for jenkins to put the db dumps in.

https://trello.com/c/0jAgMP0d/405-gdrive-client-alternatives-db-dumps-to-s3-not-drive

### DIFF
--- a/terraform/modules/iam-users/developers.tf
+++ b/terraform/modules/iam-users/developers.tf
@@ -93,7 +93,7 @@ resource "aws_iam_group_policy_attachment" "dev_s3_only_iam_manage_account" {
   policy_arn = "${var.iam_manage_account_policy_arn}"
 }
 
-resource "aws_iam_policy" "dev_uploads_s3_access" {
+resource "aws_iam_policy" "dev_s3_access" {
   name = "devUploadsS3Access"
 
   policy = <<POLICY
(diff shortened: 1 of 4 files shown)


terraform/modules/iam-users/developers.tf
```diff
@@ -93,7 +93,7 @@ resource "aws_iam_group_policy_attachment" "dev_s3_only_iam_manage_account" {
   policy_arn = "${var.iam_manage_account_policy_arn}"
 }
 
-resource "aws_iam_policy" "dev_uploads_s3_access" {
+resource "aws_iam_policy" "dev_s3_access" {
   name = "devUploadsS3Access"
 
   policy = <<POLICY
```
katstevens: Is the plan to move this inline policy to its own `aws_iam_policy_document` at some point as well?

benvand: Yeah, wanted to avoid the PR noise. I actually put in a ticket on the tech debt board this morning to have a look at some of this stuff. 
Problem is converting this one now means I should probably convert the jenkins one too, and that's a beast.

PR #508, opened 2019-03-05
app environment variables: use ugly "if" statement to switch these to the .apps.internal addresses for preview only

risicle:
https://trello.com/c/H53K45It

I apologize for the temporary ugliness of this solution, but it really is meant to be temporary as I switch the environments over to using internal urls one by one.


risicle:
I've tweaked and tested this and it appears to actually work.

PR #513, opened 2019-03-06
[TAKE 2] app environment variables: use ugly "if" statement to switch these to  the .apps.internal addresses for preview only

risicle:
https://trello.com/c/H53K45It

This is a re-attempt at https://github.com/alphagov/digitalmarketplace-aws/pull/508 which actually includes the final fix it needed to work.

### DIFF
--- a/paas/admin-frontend.j2
+++ b/paas/admin-frontend.j2
@@ -5,7 +5,11 @@
       AWS_SECRET_ACCESS_KEY: {{ aws_secret_access_key }}
 
       DM_DATA_API_AUTH_TOKEN: {{ api.auth_tokens[0] }}
+  {% if environment == "preview" %}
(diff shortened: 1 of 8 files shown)


lfdebrux:
I can't see any difference between this and #508, have I misunderstood something?

risicle:
Ahhh it's certainly subtle, but #508 feeds `DM_DATA_API_URL` to the router app but this one feeds `DM_API_URL` to it instead.

risicle:
Oh actually that's interesting... the previous merge *wasn't* using the wrong env name... wonder what did it then...

risicle:
Ok I'm going to merge this simply so I can take a better look at the running instance if/when it's going wrong rather than immediately panicking and reverting.

PR #517, opened 2019-03-11
Convert json to policy docs in cloudtrail/cloudtrail-cloudwatch module

benvand:
Wanted to test whether you can use an IAM policy as an assume_role_policy
You cannot:
https://www.terraform.io/docs/providers/aws/r/iam_role.html#assume_role_policy

But thought we might as well commit this.

It changes the name of the `module.cloudtrail-cloudwatch.aws_iam_role.cloud_watch_logs_role` to `module.cloudtrail-cloudwatch.aws_iam_role.cloudtrail_to_cloudwatch_role` so we'd have to do a `tf state mv` for that, but otherwise just `plan` and `apply`

### DIFF
--- a/terraform/modules/cloudtrail/cloudtrail-cloudwatch/main.tf
+++ b/terraform/modules/cloudtrail/cloudtrail-cloudwatch/main.tf
@@ -3,53 +3,50 @@ resource "aws_cloudwatch_log_group" "cloudtrail_log_group" {
   retention_in_days = "${var.retention_in_days}"
 }
 
-resource "aws_iam_role" "cloud_watch_logs_role" {
+# When this policy is attached to a role as an assume_role_policy the aws cloudtrail service can assume the role
+data "aws_iam_policy_document" "cloudtrail_assume_role_policy_document" {
+  statement {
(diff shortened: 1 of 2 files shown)


risicle:
Yeah I always wondered why we used inline json for these sections...

katstevens:
So... let me see if I've got this right. Previously, we had two resources
- `"aws_iam_role" "cloud_watch_logs_role"`, with an inline policy to assume role
- `"aws_iam_role_policy" "cloud_watch_policy"` with an inline policy to stream logs to cloudwatch

Now we have three resources:
- `"aws_iam_role" "cloudtrail_to_cloudwatch_role"`, which links to a separate assume role policy document
- `"aws_iam_policy" "create_cloudtrail_logs_in_cloudwatch_cloudtrail_log_group_policy"`, which links to a separate stream logs policy document
- `"aws_iam_role_policy_attachment" "attach_create_logs_in_log_group_to_cloudtrail_to_cloudwatch_role"`, which... links the log streaming role to the switch role policy? Is that right?

benvand:
@katstevens Exactly right 💯

PR #520, opened 2019-03-18
Add versioning to cleaned_db_dumps bucket

jonodrew:
[Ticket](https://trello.com/c/FDv4IfBt/417-clean-prod-db-dump-and-apply-deletes-the-old-dev-dump-and-creates-the-new-dump-we-could-just-use-versions-for-this)

Using the version capabilities of S3 buckets we can have old versions of the database dump, just in case we need them. At the moment we just delete them.

### DIFF
--- a/scripts/apply-cleaned-db-to-target.sh
+++ b/scripts/apply-cleaned-db-to-target.sh
@@ -4,13 +4,12 @@ set -euo pipefail
 echo "Target is: ${TARGET}"
 
 mkdir -p ./dumps
-LATEST_PROD_DUMP=$(aws s3 ls digitalmarketplace-database-backups | grep production | sort -r | head -1 | awk '{print $4}')
-pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > ./dumps/cleaned-"${LATEST_PROD_DUMP%.*}"
+pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > cleaned-db-dump.sql.gz
 
 if [ "${TARGET}" == 's3' ]; then
   echo 'Uploading cleaned dump to S3'
 elif [ "${TARGET}" == 'preview' ] || [ "${TARGET}" == 'staging' ]; then
-  echo "Migrating cleaned dump to ${TARGET} and uploading to Google Drive"
+  echo "Migrating cleaned dump to ${TARGET} and uploading to S3 bucket"
   cf target -s ${TARGET}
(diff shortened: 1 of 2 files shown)


scripts/apply-cleaned-db-to-target.sh
```diff
@@ -5,12 +5,12 @@ echo "Target is: ${TARGET}"
...
 mkdir -p ./dumps
 LATEST_PROD_DUMP=$(aws s3 ls digitalmarketplace-database-backups | grep production | sort -r | head -1 | awk '{print $4}')
-pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > ./dumps/cleaned-"${LATEST_PROD_DUMP%.*}"
+pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > ./dumps/cleaned-"${LATEST_PROD_DUMP}"
 
 if [ "${TARGET}" == 's3' ]; then
   echo 'Uploading cleaned dump to S3'
 elif [ "${TARGET}" == 'preview' ] || [ "${TARGET}" == 'staging' ]; then
-  echo "Migrating cleaned dump to ${TARGET} and uploading to Google Drive"
+  echo "Migrating cleaned dump to ${TARGET} and uploading to S3 buckete"
```
benvand: > buckete

benvand: > buckete

scripts/apply-cleaned-db-to-target.sh
```diff
@@ -5,12 +5,12 @@ echo "Target is: ${TARGET}"
 
 mkdir -p ./dumps
 LATEST_PROD_DUMP=$(aws s3 ls digitalmarketplace-database-backups | grep production | sort -r | head -1 | awk '{print $4}')
-pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > ./dumps/cleaned-"${LATEST_PROD_DUMP%.*}"
+pg_dump --no-acl --no-owner --clean postgres://postgres:@localhost:63306/postgres | gzip > ./dumps/cleaned-"${LATEST_PROD_DUMP}"
```
benvand: On jenkins:
```
> aws s3 ls digitalmarketplace-database-backups | grep production | sort -r | head -1 | awk '{print $4}'
production-201903190300.sql.gz.gpg
```

The file is brought down from S3 with the timestamp attached. 
The `%.*` strips the `.gpg` suffix so it becomes:
```
production-201903190300.sql.gz.gpg
```

This means we'll be uploading the file as `production-201903190300.sql.gz` which won't action the versioning as it'll have a different name in the bucket.

PR #521, opened 2019-03-18
preview private routes: open a public .cloudapps.digital route for /_metrics only

risicle:
https://trello.com/c/QOj2iMLO

Even though we're likely to yank this all from preview soon, it would be nice to have committed the proper solution for this so it's clear what needs to be restored to bring private apps back...

### DIFF
--- a/vars/preview.yml
+++ b/vars/preview.yml
@@ -4,6 +4,7 @@ maintenance_mode: live
 
 api:
   routes:
+    - dm-api-{env}.cloudapps.digital/_metrics
     - dm-api-{env}.apps.internal
   memory: 2GB
   egress_to_applications:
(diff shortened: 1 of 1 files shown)


lfdebrux:
I'm going to rebase and merge this while @risicle is away

risicle:
Once merged, you may want to "back out" all the changes and bring preview back to using non-private routes. This consists of 

 - the changes in this vars file, which for preview can go back to being very minimal.
 - the `{% if environment == 'preview' %}` branches in the manifest templates that set the envs for the instances.

Think a little bit about the order you do these in if you want to make this change "seamless" - you probably want to do the env change before the vars and perform a "release all apps" in between. Just think about it a bit - it's not rocket science.

PR #522, opened 2019-03-20
413-re-create-an-alarm-for-dropped-av-sns

benvand:
https://trello.com/c/gHYwVkSu/413-re-create-an-alarm-for-dropped-av-sns

* Update comment with more info
* Update comment with correct second value
* Update readme with alarm_recovery_email_topic_arn arg for log examples
* Add new dropped-av-sns alarm module for creating dropped av alarm
* Add readme entry for new dropped av sns alarm
* Apply new dropped av sns alarm to preview
* Apply new dropped av sns alarm to production

### DIFF
--- a/terraform/modules/logging/alarms/README.md
+++ b/terraform/modules/logging/alarms/README.md
@@ -39,6 +39,7 @@ module "missing_logs_alarms" {
   environment           = "preview"
   app_names             = ["buyer-frontend", "api", "something else", "etc..."]
   alarm_email_topic_arn = "${module.email_alarm_sns.alarm_email_topic_arn}"
+  alarm_recovery_email_topic_arn = "${module.alarm_recovery_email_sns.email_topic_arn}"
 }
 ```
 
@@ -54,6 +55,7 @@ module "slow_requests_alarms" {
   environment           = "preview"
   app_name              = "router"
   alarm_email_topic_arn = "${module.email_alarm_sns.alarm_email_topic_arn}"
+  alarm_recovery_email_topic_arn = "${module.alarm_recovery_email_sns.email_topic_arn}"
 }
 ```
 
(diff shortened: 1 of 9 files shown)


terraform/modules/logging/alarms/dropped-av-sns/alarms.tf
```diff
@@ -0,0 +1,21 @@
+resource "aws_cloudwatch_metric_alarm" "status_code_alarm" {
```
katstevens: Should the `status_code_alarm` name be `dropped_av_sns_alarm`?

terraform/modules/logging/alarms/dropped-av-sns/alarms.tf
```diff
@@ -0,0 +1,21 @@
...
+  evaluation_periods = "1"
+  period             = "60"
+
+  // If totals 1 or higher
+  statistic           = "Sum"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  threshold           = "1"
+
+  // Email slack
+  alarm_actions = ["${var.alarm_email_topic_arn}"]
```
katstevens: The missing logs alarm has an empty `insufficient_data_actions` in this section, do we need it here?

terraform/modules/logging/alarms/README.md
```diff
@@ -73,4 +73,19 @@ module "router_500_alarm" {
...
+```
+
+### dropped-av-sns
+
+This module creates a single alarm that is triggered on an entry in the ${var.environment}-dropped-antivirus-sns metric.
+
+It relies on the [`${var.environment}-dropped-antivirus-sns` metrics](https://github.com/alphagov/digitalmarketplace-aws/blob/2df2d21ea8c8bd0da78a37c7f6ce3d71889d00e2/terraform/modules/logging/log-metric-filters/main.tf#L325).
+
+```
+module "router_500_alarm" {
```
katstevens: `router_500_alarm` - c+p-ed from above I think?

terraform/modules/logging/alarms/dropped-av-sns/alarms.tf
benvand: Na, we don't need i there either to be fair.
You have reminded me to put a treat_missing_data value on these though. We don't care if data is missing because we have the `missing logs` alarms

benvand: Have removed it and added missing data actions in another commit

PR #528, opened 2019-03-29
427-put-metrics-on-api

benvand:
https://trello.com/c/wGJZQFxO/427-put-metrics-on-api

* Move _metrics PaaS env arguments from brief-responses only to global base so they are available for every deployment
* Move metrics routes definitions from preview vars to common vars so these routes are defined on next release of all stages

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -10,12 +10,14 @@ router:
 api:
   routes:
     - dm-api-{env}.cloudapps.digital
+    - dm-api-{env}.cloudapps.digital/_metrics
   services:
     - digitalmarketplace_api_db
 
 search-api:
   routes:
     - dm-search-api-{env}.cloudapps.digital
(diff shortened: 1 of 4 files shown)


vars/common.yml
```diff
@@ -10,19 +10,22 @@ router:
...
     - dm-search-api-{env}.cloudapps.digital
+    - dm-search-api-{env}.cloudapps.digital/_metrics
   services:
     - search_api_elasticsearch
 
 antivirus-api:
   memory: 2G
   routes:
     - dm-antivirus-api-{env}.cloudapps.digital
+    - dm-antivirus-api-{env}.cloudapps.digital/_metrics
```
katstevens: Are we just doing the APIs first, or do we need this added to the other FEs too?

benvand: Just apis for now

PR #534, opened 2019-04-04
449-reorganise-jenkins-terraform-to-make-rebuild-easier

benvand:
https://trello.com/c/NFFmr3C7/449-reorganise-jenkins-terraform-to-make-rebuild-easier

* Split creation of jenkins/ elb and the logs bucket into separate sub modules
  * Reorganise jenkins terraform to make rebuild easier
  * Can make it so logs are consistently piped to the same bucket even in a rebuild situation by splitting out the creation of the S3 bucket from the creation of jenkins and its ELB
* Update jenkins module calls to use new format


benvand:
The make plan shows:

Screen Shot 2019-04-04 at 12 33 16


1 change
1 add
1 remove

Once this is merged I'll 'rename' the `module.jenkins.aws_s3_bucket.jenkins_logs_bucket` to `module.jenkins_elb_log_bucket.aws_s3_bucket.jenkins_logs_bucket` which will nullify the addition/ removal

The change is just moving from a semi-hardcoded name to a variable, which is what we want

terraform/modules/jenkins/log_bucket/variables.tf
```diff
@@ -0,0 +1 @@
+variable "name" {}
```
katstevens: So is this `name` the thing we supply when we define `module "jenkins_elb_log_bucket"` in `main.tf`?

benvand: Yup

PR #535, opened 2019-04-05
Change deploy-app to push new app without route

lfdebrux:
Our deploy-app recipe will create a `-release` app with a route and only removes the route after it has been started. If the app start fails this means there can be a route pointing to a dead app. See https://trello.com/c/x76uyB81/692-a-failed-api-release-will-leave-a-dangling-route for info on how this came up.

This commit changes the deploy-app recipe so that the release app will be created without a route, and then if the app fails to start nothing else will be broken 🤞. The route mapping is done by a new script, `scripts/map-route.sh`, that gets the desired route(s) from the app manifest.

This should also have the bonus effect of meaning our release process causes less downtime in our metrics.

This has been tested on preview.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ deploy-app: ## Deploys the app to PaaS
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
 	$(if ${RELEASE_NAME},,$(error Must specify RELEASE_NAME))
 	$(if ${STAGE},,$(error Must specify STAGE))
-	cf push --no-start -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME}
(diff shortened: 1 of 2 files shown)


katstevens:
So to check my understanding, I think this is what's now happening:

- create the release app without a route, don't start it
- add the network policy ingress/egress where necessary
- sleep 60
- start the release app
- add the route to the release app (new step)
- unmap the route for the old app
- sleep 60
- stop the old app
- delete the old app
- rename the release app 

I think our sleeps are still in the right places? Is there anything missing?

lfdebrux:
> So to check my understanding, I think this is what's now happening:
> 
> * create the release app without a route, don't start it
> * add the network policy ingress/egress where necessary
> * sleep 60
> * start the release app
> * add the route to the release app (new step)
> * unmap the route for the old app
> * sleep 60
> * stop the old app
> * delete the old app
> * rename the release app
> 
> I think our sleeps are still in the right places? Is there anything missing?

I just tested the map-route command, it's pretty quick.

> Let's try it - I can merge my API PR straight after this is merged.

👍

PR #545, opened 2019-05-09
Add new add-ip-whitelist-route-service.sh script

benvand:
* Add new variable file for script to use.
  * These cannot be part of the manifest bacuse the manifest template variables are filled in inside generate-manifest which will also strip any custom variables out
* Add script which takes uses variables file to add routes to ip-whitelisting-service
* Run script in deployment after routes are created

https://trello.com/c/Ktinf9rQ/457-script-to-add-ip-whitelisted-routes-on-deploy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,10 @@ deploy-app: ## Deploys the app to PaaS
 
 	# Create the route for the release app
 	./scripts/map-route.sh ${APPLICATION_NAME}-release <(make -s -C ${CURDIR} generate-manifest)
+
+	# Make sure relevant routes are whitelisted (can be found in vars/ip_whitelist_routes.json)
+	.scripts/add-ip-whitelist-route-service.sh ${APPLICATION_NAME}
+
 	# Delete the route for the old app
 	@if cf app ${APPLICATION_NAME} >/dev/null; then ./scripts/unmap-route.sh ${APPLICATION_NAME}; fi
 
--- a/vars/ip_whitelist_routes.json
+++ b/vars/ip_whitelist_routes.json
@@ -1,0 +1,35 @@
+{
+  "preview": [
+    {"hostname": "dm-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-search-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-antivirus-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-preview", "path": "/_metrics"},
+    {"hostname": "dm-preview", "path": "/admin/_metrics"},
+    {"hostname": "dm-preview", "path": "/buyers/_metrics"},
+    {"hostname": "dm-preview", "path": "/suppliers/_metrics"},
+    {"hostname": "dm-preview", "path": "/suppliers/opportunities/_metrics"},
+    {"hostname": "dm-preview", "path": "/user/_metrics"}
+  ],
+  "staging": [
+    {"hostname": "dm-api-staging", "path": "/_metrics"},
+    {"hostname": "dm-search-api-staging", "path": "/_metrics"},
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -126,6 +126,9 @@ deploy-app: ## Deploys the app to PaaS
 	# Delete the route for the old app
 	@if cf app ${APPLICATION_NAME} >/dev/null; then ./scripts/unmap-route.sh ${APPLICATION_NAME}; fi
 
+	# Make sure relevant routes are whitelisted (can be found in vars/ip_whitelist_routes.json)
+	.scripts/add-ip-whitelist-route-service.sh ${APPLICATION_NAME}
```
katstevens: Should the new route be whitelisted before we unmap the old route?

vars/ip_whitelist_routes.json
```diff
@@ -0,0 +1,23 @@
+{
+  "preview": [
+    {"hostname": "dm-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-search-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-antivirus-api-preview", "path": "/_metrics"},
+    {"hostname": "dm-preview", "path": "/suppliers/_metrics"},
+    {"hostname": "dm-preview", "path": "/suppliers/opportunities/_metrics"}
```
katstevens: Is there a programmatic way to do this so that we don't have to update this later when we add `/_metrics` to the other FE apps?

We can get a list of hostnames and paths for an app from `cf curl /v2/apps/${APP_GUID}/routes | jq -r '.resources[].entity.host'` and `cf curl /v2/apps/${APP_GUID}/routes | jq -r '.resources[].entity.path'` (we use this method on the `map-route.sh` and `unmap-route.sh` scripts).

benvand: While I don't like the idea of having to have another variables file just for this one script I also don't like the idea of hard coding `_metrics` in to everything. We could conceivably change `_metrics` to anything we like. Equally we could want to IP whitelist a route that is not `_metrics`.

It's a bit of a rock and a hard place solution. But I erred on the side of explicitly defining them.

lfdebrux: Is this info not already in the app manifest?

benvand: No

1 There is no way of telling between routes we want protected from the manifest and routes we don't
2 The routes exist in the vars files as uninterpolated templates  eg `dm-{env}.apps.internal`

lfdebrux: We could add that information to the manifest? Rather than adding a new source of information. Then pass the expanded manifest to the script?

benvand: I tried that, but they're for fundamentally different information.

We need to specify a subset of the routes that are defined in the manifest. _But_ we need to have them once they've been interpolated with the various variables.

vars/ip_whitelist_routes.json
```diff
@@ -0,0 +1,23 @@
...
+    {"hostname": "dm-staging", "path": "/suppliers/opportunities/_metrics"}
+  ],
+  "production": [
+    {"hostname": "dm-api-production", "path": "/_metrics"},
+    {"hostname": "dm-search-api-production", "path": "/_metrics"},
+    {"hostname": "dm-antivirus-api-production", "path": "/_metrics"},
+    {"hostname": "dm-production", "path": "/suppliers/_metrics"},
+    {"hostname": "dm-production", "path": "/suppliers/opportunities/_metrics"}
+  ]
+}
```
lfdebrux:
```suggestion
}

```

Makefile
benvand: I _think_ it will be fine to do it before we unmap actually yeah. I've updated and I'm going to give it a whirl on preview

PR #546, opened 2019-05-13
extra conditions for slow request alarms

agalamatis:
trello: https://trello.com/c/Lgu5r2Rr/773-exclude-slow-requests-over-a-certain-size-from-triggering-alarms

extra conditions are:
- request begins with POST
- size larger than 500k

### DIFF
--- a/terraform/modules/logging/log-metric-filters/main.tf
+++ b/terraform/modules/logging/log-metric-filters/main.tf
@@ -272,7 +272,7 @@ resource "aws_cloudwatch_log_metric_filter" "request_time_bucket_router_7" {
 
 resource "aws_cloudwatch_log_metric_filter" "request_time_bucket_router_8" {
   name           = "${var.environment}-router-request-times-8"
-  pattern        = "{$$.requestTime >= 5 && $$.requestTime < 10 && $$.request != \"*/_status?ignore-dependencies *\" && $$.request != \"POST*\" && $$.requestSize < 500000}"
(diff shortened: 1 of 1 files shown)


terraform/modules/logging/log-metric-filters/main.tf
```diff
@@ -285,7 +285,7 @@ resource "aws_cloudwatch_log_metric_filter" "request_time_bucket_router_8" {
 
 resource "aws_cloudwatch_log_metric_filter" "request_time_bucket_router_9" {
   name           = "${var.environment}-router-request-times-9"
-  pattern        = "{$$.requestTime >= 10 && $$.request != \"*/_status?ignore-dependencies *\"}"
+  pattern        = "{$$.requestTime >= 10 && $$.request != \"*/_status?ignore-dependencies *\" && $$.request != \"POST*\" && $$.size < 500000}"
```
katstevens: Will this include a slow POST with a small request size?

agalamatis: all conditions operated by &&, so all four must be true to send the alarm

katstevens: I think we probably want to know about small slow POSTs. I'd be happy to remove the `!= POST` condition.

agalamatis: I can also separate the two conditions with an || and put them into brackets.

PR #550, opened 2019-05-16
enabling validation

agalamatis:
trello: https://trello.com/c/39BmiSKF/500-enable-cloud-trail-log-file-validation

documentation: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-log-file-validation-intro.html


lfdebrux:
Well that was easy! :+1: Do we want to put some extra security on the hash files?

agalamatis:
> Well that was easy! 👍 Do we want to put some extra security on the hash files?

I don't think it's necessary, because the validation will fail if either the log or its digest is even moved, let alone been tampered with.

But we should create a new ticket for the validation task as I mentioned in the trello card.

PR #551, opened 2019-05-16
Increase DB backup app memory

katstevens:
Trello: https://trello.com/c/f1PbVXEW/785-db-backup-app-runs-out-of-memory

Our `db-backup` app is created with 128MB of memory, even though our maximum memory usage for the task is later set to 2G:

`cf run-task db-backup "/app/create-db-dump.sh" --name db-backup -m 2G -k 2G`

Also adds a commit to hide the DB service credentials from Jenkins console output (please check I've got my bash the right way round here...)

### DIFF
--- a/db-backup/create-db-dump.sh
+++ b/db-backup/create-db-dump.sh
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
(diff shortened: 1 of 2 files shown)


katstevens:
Thanks all 👍

PR #557, opened 2019-05-17
457-script-to-add-ip-whitelisted-routes-on-deploy-4

benvand:
* Do not use reserved var name PATH


scripts/add-ip-whitelist-route-service.sh
```diff
@@ -11,8 +11,8 @@ STAGE=$( echo $1 | awk '{ print tolower($0) }' )
 
 IFS=$'\n'
 for HOSTNAME_PATH in $(jq -r ".${STAGE}[] | \"\(.hostname) \(.path)\"" vars/ip_whitelist_routes.json); do
```
lfdebrux:
```suggestion
for HOSTNAME_PATH in $(jq -r '.[env.STAGE][] | "\(.hostname) \(.path)"' vars/ip_whitelist_routes.json); do
```

PR #564, opened 2019-05-22
paas instances: scale down production instances following G11's close

risicle:
Both number of instances and memory

### DIFF
--- a/vars/production.yml
+++ b/vars/production.yml
@@ -4,7 +4,7 @@ maintenance_mode: live
 instances: 5
 
 router:
-  instances: 5
+  instances: 3
   rate_limiting_enabled: enabled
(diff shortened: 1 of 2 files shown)


vars/production.yml
```diff
@@ -15,10 +15,10 @@ router:
 
 api:
   memory: 2GB
-  instances: 20
+  instances: 10
 
 buyer-frontend:
-  instances: 10
+  instances: 8
```
katstevens: I thought we were on the default (5) for buyer before? Can we just delete this bit?

risicle: If ya like

risicle:
Also dropped the # of router instances

PR #565, opened 2019-05-23
S3 bucket policies and versioning

agalamatis:
trello: https://trello.com/c/w4tnuB26/505-enable-s3-bucket-versioning
also: https://trello.com/c/kyFOBl0M/512-allow-only-encrypted-connection-to-s3-buckets

### DIFF
--- a/terraform/environments/preview/s3_buckets.tf
+++ b/terraform/environments/preview/s3_buckets.tf
@@ -8,6 +8,29 @@ resource "aws_s3_bucket" "server_access_logs_bucket" {
   }
 }
 
+resource "aws_s3_bucket_policy" "server_access_logs_bucket" {
+  bucket = "${aws_s3_bucket.server_access_logs_bucket.id}"
+
+  policy = <<POLICY
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Deny",
+      "Principal": "*",
+      "Action": "*",
+      "Resource": "arn:aws:s3:::digitalmarketplace-logs-preview-preview/*",
+      "Condition": {
+        "Bool": {
+          "aws:SecureTransport": "false"
+        }
+      }
+    }
+  ]
+}
+POLICY
+}
+
 # TODO remove these hard-coded definitions in favour of using the terraform/modules/s3-document-bucket module after the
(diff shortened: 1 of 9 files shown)


terraform/environments/preview/s3_buckets.tf
```diff
@@ -306,9 +381,57 @@ resource "aws_s3_bucket" "g7-draft-documents_bucket" {
   }
 }
 
+resource "aws_s3_bucket_policy" "g7-draft-documents_bucket" {
+  bucket = "${aws_s3_bucket.g7-draft-documents_bucket.id}"
+
+  policy = <<POLICY
```
katstevens: Is there a reason we're using an inline policy here over a `policy_document`? I think we want to use the latter where possible.

agalamatis: I was following the convention of existing policies, if they were inlined I left them inlined. But I can change them, if everyone prefers the TF syntax over inlined JSON.

katstevens:
To check my understanding - are these `Deny` policies intended to disable everything on the baseline set of permissions, and we then add specific `Allow` policies on top of that? How do we manage buckets that only have a `Deny` policy?

(I'd definitely squash the terraformat linting commits, by the way.)

agalamatis:
> To check my understanding - are these `Deny` policies intended to disable everything on the baseline set of permissions, and we then add specific `Allow` policies on top of that? How do we manage buckets that only have a `Deny` policy?

Yes, from what I understand all statements work in conjunction. Normally list bucket is denied to all, therefore we need to allow it to certain users. Non-encrypted connections are allowed, therefore we need to deny it.

> (I'd definitely squash the terraformat linting commits, by the way.)

I will

katstevens:
@agalamatis OK, that makes sense. What happens when we add the `Deny` policy to say, the cross region replication buckets, which currently don't have any `Allow` policies set?

agalamatis:
> @agalamatis OK, that makes sense. What happens when we add the `Deny` policy to say, the cross region replication buckets, which currently don't have any `Allow` policies set?

@katstevens The statements of a policy change the default behaviour of buckets. A new bucket accepts both HTTP and HTTPS, so I added a statement that denies non-encrypted connections. You think that this will have some kind of side-effect?

PR #569, opened 2019-05-31
paas manifests: hook up /_metrics route for router

risicle:
~This would allow https://github.com/alphagov/digitalmarketplace-router/pull/64 to be implemented in a slightly neater way.~

To enable https://trello.com/c/wkDN6HHk

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -1,11 +1,15 @@
 ---
 router:
   routes:
+    - dm-router-{env}.cloudapps.digital
+    - dm-router-{env}.cloudapps.digital/_metrics
     - www.{env}.marketplace.team
     - api.{env}.marketplace.team
     - search-api.{env}.marketplace.team
     - antivirus-api.{env}.marketplace.team
     - assets.{env}.marketplace.team
+  services:
+    - digitalmarketplace_prometheus
 
 api:
   routes:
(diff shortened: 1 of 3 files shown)


vars/common.yml
```diff
@@ -1,11 +1,15 @@
 ---
 router:
   routes:
+    - dm-router-{env}.cloudapps.digital
+    - dm-router-{env}.cloudapps.digital/_metrics
     - www.{env}.marketplace.team
```
benvand: This will get overwritten on production by the router block in 👍 
https://github.com/alphagov/digitalmarketplace-aws/blob/master/vars/production.yml#L9

https://github.com/alphagov/digitalmarketplace-aws/blob/master/vars/README.md
> Values in a STAGE named file will override those in common.yml.

benvand: This will get overwritten on production by the router block in:
https://github.com/alphagov/digitalmarketplace-aws/blob/master/vars/production.yml#L9

https://github.com/alphagov/digitalmarketplace-aws/blob/master/vars/README.md
> Values in a STAGE named file will override those in common.yml.

risicle: Ah right yes - I was aware this PR was going to need some further screwing with, but hadn't noticed that one... had only worried about getting preview working to start with.

PR #572, opened 2019-05-31
Add role for validating CloudWatch logs

lfdebrux:
We want to be able to validate our CloudWatch logs; this PR adds a role with the correct access rights. Adding it to the existing cloudtrail module means that it will be automatically picked up by all the accounts.

It has been configured so that the role in each account can be assumed by the main account; this is so that we can validate the logs periodically with Jenkins.

### DIFF
--- a/terraform/modules/cloudtrail/main.tf
+++ b/terraform/modules/cloudtrail/main.tf
@@ -9,6 +9,12 @@ module "cloudtrail-cloudwatch" {
   retention_in_days = 731                               // As per RE convention on keeping logs for 2 years
 }
 
+module "cloudtrail-validate-logs-role" {
+  source          = "./modules/cloudtrail-validate-logs-role"
+  assume_role_arn = "arn:aws:iam::${var.validate_account_id}:root"
+  s3_bucket_arn   = "${module.cloudtrail-bucket.s3_bucket_arn}"
+}
+
 module "cloudtrail-cloudtrail" {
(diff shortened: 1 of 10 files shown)


lfdebrux:
[log file validation job](https://trello.com/c/Ijzuk0qY/508-log-file-validation-job)

terraform/modules/cloudtrail/modules/cloudtrail-validate-logs-role/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "cloudtrail_validate_logs_role_arn" {
```
benvand: Unless this is used anywhere I wouldn't bother defining it

terraform/modules/cloudtrail/main.tf
```diff
@@ -9,11 +9,17 @@ module "cloudtrail-cloudwatch" {
   retention_in_days = 731                               // As per RE convention on keeping logs for 2 years
 }
 
+module "cloudtrail-validate-logs-role" {
+  source          = "./modules/cloudtrail-validate-logs-role"
+  assume_role_arn = "${var.validate_account_id}"
```
benvand: hard to know what `assume_role_arn` is at a glance. It's called `account_id`, `validate_account_id` and `assume_role_arn` in different places. Can we normalise this?

PR #573, opened 2019-06-03
NOTICKET-make-ip-whitelist-script-more-robust

benvand:
Add APPLICATION_NAME arg to ip-whitelist script so script only adds route relevant to current deployment

This script may cause failures when routes do not yet exist. So if a new route is added to the vars here for whitelisting, and this script runs without the route existing already, we will see deployment failures.

* Add new variable to script
* Supply new var in Makefile
* Nest values in json by app name


risicle:
Is there an easy way we could make the `APPLICATION_NAME` optional? So that the script just substituted the value with the jq equivalent of `*` for example.

benvand:
> Is there an easy way we could make the APPLICATION_NAME optional?

Not that I can find, substituting for null in jq is not easy.

risicle:
Not important

PR #578, opened 2019-06-10
digitalmarketplace not digitalmarketplaces

lfdebrux:


katstevens:
I thought we'd decided not to do this?

lfdebrux:
> I thought we'd decided not to do this?

Ah, I thought we had just agreed to not worry about merging buckets/log groups

benvand:
@katstevens @lfdebrux 
We can go either way, but maybe we should do it actually. Then we can move the logs from the badly named buckets into the proper buckets and delete the log groups. The logs will still exist just not in cloudwatch.

That way we don't need to keep track of duplicate buckets (either in or out of terraform (both would be a bit rubbish))

benvand:
I'll approve pending comment from @katstevens

katstevens:
Yep, that sounds sensible to me!

lfdebrux:
Terraform has been applied and logs in buckets migrated

PR #579, opened 2019-06-10
Make it harder to accidentally destroy a Jenkins instance

lfdebrux:
Don't be like me ;)

### DIFF
--- a/terraform/modules/jenkins/jenkins/ec2.tf
+++ b/terraform/modules/jenkins/jenkins/ec2.tf
@@ -3,11 +3,12 @@ provider "aws" {
 }
 
 resource "aws_instance" "jenkins" {
-  ami                    = "${var.ami_id}"
-  instance_type          = "${var.instance_type}"
-  iam_instance_profile   = "${var.jenkins_instance_profile}"
-  key_name               = "${aws_key_pair.jenkins.key_name}"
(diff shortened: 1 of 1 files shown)


terraform/modules/jenkins/jenkins/ec2.tf
```diff
@@ -4,6 +4,7 @@ provider "aws" {
 
 resource "aws_instance" "jenkins" {
   ami                    = "${var.ami_id}"
+  disable_api_termination= true  # prevent this instance from being destroyed from the console
```
katstevens: Does this also prevent us destroying the instance with Terraform? Would we have to change & apply this attribute first if we needed to destroy for some reason?

lfdebrux: I believe if for some reason we wanted to destroy the instance we would just remove it from the terraform file, and then terraform would see that it was terminated.

Alternatively we could terminate it from the AWS console by first removing the termination protection from the AWS console (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingDisableAPITermination). This information is in the terraform manual for the aws_ec2_instance resource.

katstevens: Hmm, ok. Let's make sure we note this down in the manual.

PR #581, opened 2019-06-12
Cyber audit support permissions

danjoneslf:
Switch security audit role commit ref to apply support:* permissions for Support/Trusted Advisor.


danjoneslf:
I noticed this PR was marked as behind so I've just rebased from master.

PR #589, opened 2019-07-12
gpg2 output and required args changed with 2.x

benvand:
* Allow pinentry mode
* Get only parent secret key (not child)


lfdebrux:
Does this mean we can get rid of the gpg config file from the Ansible playbook?

benvand:
No afraid not, `allow-loopback-pinentry` has to be enabled there, and on the command line and the `--yes` argument is required to not make it ask for confirmation... Seems like a bit much to me but 🤷‍♂

PR #595, opened 2019-08-02
add log metric and alarm for login.reset-email.bad-role

risicle:
https://trello.com/c/Vxmq7W0m

Seems to work.

Alarm is only enabled for production, because I don't think it's a particular worry for other environments, where we can still view the metric from the aws console for testing & debugging purposes.


risicle:
Probably should do.

PR #603, opened 2019-08-28
1071-why-did-api-release-to-preview-fail

benvand:
## Add health-check-http-endpoint none when creating db migration app

Recieved error:

Health check type must be 'http' to set a health check HTTP endpoint.

Assumption is that we cannot set health-check-http-endpoint endpoint value when setting health-check-type to none. Command line arguments override manifest values so this should solve the problem.

Screen Shot 2019-08-28 at 16 48 15

https://trello.com/c/IgxiQyEE/1071-why-did-api-release-to-preview-fail

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
 .PHONY: deploy-db-migration
 deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
-	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none -i 1 -m 128M -c 'sleep 2h'
+	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none --health-check-http-endpoint none -i 1 -m 128M -c 'sleep 2h'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
 .PHONY: deploy-db-migration
 deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
-	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none -i 1 -m 128M -c 'sleep 2h'
+	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none --health-check-http-endpoint none -i 1 -m 128M -c 'sleep 2h'
```
katstevens: If `--health-check-type none` is present, why do we need `--health-check-http-endpoint none`?

benvand: I don't know, not a great design. But reading between the lines of the error

> Health check type must be 'http' to set a health check HTTP endpoint.

we're setting `--health-check-type none` with the command line override then trying to set `health-check-http-endpoint` in the manifest here: https://github.com/alphagov/digitalmarketplace-aws/blob/master/paas/_base.j2#L18

katstevens: Ahh I see - we disable healthcheck on `cf push` but then `generate-manifest` puts one in anyway? 🤔

benvand: Yeah, the manual puts in a url for the health check even though we set the 'healthcheck type' to `none`. CF throws a wobbly :/

PR #604, opened 2019-08-29
1071-why-did-api-release-to-preview-fail-2

benvand:
Create new manifest for a no-route app with db access to perform db-migration

Instead of using the api app manifest to perform migrations we can use a separate manifest that only gives the migrations app what it needs.

This removes the need to override loads of manifest values in the call to cf push which caused the below error

63873250-2ba5e000-c9b7-11e9-8802-d0651ae51c31

https://trello.com/c/IgxiQyEE/1071-why-did-api-release-to-preview-fail

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ deploy-db-backup-app: virtualenv ## Deploys the db backup app
 	$(eval export S3_POST_URL_DATA=$(shell ${VIRTUALENV_ROOT}/bin/python ./scripts/generate-s3-post-url-data.py digitalmarketplace-database-backups ${DUMP_FILE_NAME}))
 
 	# Deploy the backup app
-	cf push db-backup -f <(make -s -C ${CURDIR} generate-manifest ARGS="-v DUMP_FILE_NAME -v S3_POST_URL_DATA") -o digitalmarketplace/db-backup
+	cf push db-backup -f <(make -s -C ${CURDIR} generate-manifest ARGS="-v DUMP_FILE_NAME -v S3_POST_URL_DATA") -o digitalmarketplace/db-backup --no-route
 	cf set-env db-backup PUBKEY "$$(cat ${DM_CREDENTIALS_REPO}/gpg/database-backups/public.key)"
 	cf restage db-backup
 
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
 .PHONY: deploy-db-migration
 deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
-	cf push ${APPLICATION_NAME}-db-migration -f <(APPLICATION_NAME=db-migration make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route
+	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest APPLICATION_NAME=db-migration) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route
 	cf stop ${APPLICATION_NAME}-db-migration
 	cf run-task ${APPLICATION_NAME}-db-migration "cd /app && venv/bin/python application.py db upgrade" --name ${APPLICATION_NAME}-db-migration
 
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
 .PHONY: deploy-db-migration
 deploy-db-migration: ## Deploys the db migration app
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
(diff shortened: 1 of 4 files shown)


paas/db-backup.j2
```diff
@@ -6,7 +6,6 @@
```
benvand: See L4

paas/db-migration.j2
```diff
@@ -0,0 +1,13 @@
+{% extends "_base.j2" %}
```
benvand: Copied from paas/db-backup.j2

vars/common.yml
```diff
@@ -85,3 +85,9 @@ db-backup:
   memory: 128M
   services:
```
benvand: Copied from db-backup

Makefile
```diff
@@ -153,7 +153,7 @@ deploy-db-backup-app: virtualenv ## Deploys the db backup app
 	$(eval export S3_POST_URL_DATA=$(shell ${VIRTUALENV_ROOT}/bin/python ./scripts/generate-s3-post-url-data.py digitalmarketplace-database-backups ${DUMP_FILE_NAME}))
 
 	# Deploy the backup app
-	cf push db-backup -f <(make -s -C ${CURDIR} generate-manifest ARGS="-v DUMP_FILE_NAME -v S3_POST_URL_DATA") -o digitalmarketplace/db-backup
+	cf push db-backup -f <(make -s -C ${CURDIR} generate-manifest ARGS="-v DUMP_FILE_NAME -v S3_POST_URL_DATA") -o digitalmarketplace/db-backup --no-route
 	cf set-env db-backup PUBKEY "$$(cat ${DM_CREDENTIALS_REPO}/gpg/database-backups/public.key)"
```
benvand: For saftey, see commit

Makefile
```diff
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
 .PHONY: deploy-db-migration
 deploy-db-migration: ## Deploys the db migration app
```
benvand: * Overrides are specified in manifest and vars files now
* `APPLICATION_NAME=db-migration` makes the generate manifest step creat with the new manifest.
* Leave no-route for safety

risicle:
So the actual app image is going to be generated from the `digitalmarketplace-api` `Dockerfile` too?

How is this all different from the previous `api-migration` app we had?

benvand:
Yeah the image is still the API dockerfile. That provides the flask - alembic deps required for the migrations. The PAAS app is generated from a different manifest file (similar to the db-backup manifest file) to avoid having to override a load of stuff on the cl when we cf push.

risicle:
AhhhhI think I get it

paas/db-migration.j2
```diff
@@ -0,0 +1,14 @@
...
+
+{% block routes %}
+    no-route: true
+{% endblock %}
+
+{% block health_check %}
+    health-check-type: none
```
lfdebrux: This should be `process`, `none` is deprecated

```suggestion
    health-check-type: process
```

Makefile
```diff
@@ -127,7 +127,7 @@ deploy-app: ## Deploys the app to PaaS
...
 	$(if ${APPLICATION_NAME},,$(error Must specify APPLICATION_NAME))
-	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route --health-check-type none --health-check-http-endpoint none -i 1 -m 128M -c 'sleep 2h'
+	cf push ${APPLICATION_NAME}-db-migration -f <(APPLICATION_NAME=db-migration make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route
```
katstevens: I'm a bit confused about using `APPLICATION_NAME=db-migration` and `${APPLICATION_NAME}:${RELEASE_NAME}` (which presumably has `APPLICATION_NAME=api` to get the right Docker image) in the same line...?

lfdebrux: This might be clearer/less prone to failure

```suggestion
	cf push ${APPLICATION_NAME}-db-migration -f <(make -s -C ${CURDIR} generate-manifest APPLICATION_NAME=db-migration) -o digitalmarketplace/${APPLICATION_NAME}:${RELEASE_NAME} --no-route
```

lfdebrux: Although to be honest, since we're special casing the api app here, we might as well take out the outer ${APPLICATION_NAME} references and hardcode api

```suggestion
	cf push api-db-migration -f <(APPLICATION_NAME=db-migration make -s -C ${CURDIR} generate-manifest) -o digitalmarketplace/api:${RELEASE_NAME} --no-route
```

benvand:
@lfdebrux good call, committed, didn't know you could pass the env var through like that.

benvand:
Or is it an argument passed to make that overrides an env var? Would look all this up if I had time.

lfdebrux:
> Or is it an argument passed to make that overrides an env var? Would look all this up if I had time. It works whatever it is.

It's a Make thing, the way envvars and Make variables (which are not the same thing) interact is very confusing

https://www.gnu.org/software/make/manual/html_node/Values.html#Values

PR #611, opened 2019-09-25
Disable autoescape for jinja templates

jonodrew:
The jinja templates are solely used for paas manifest configuration. We're not sure what the impact of autoescaping will be, either now or in the future, and so we've taken a positive decision to turn it off.

I've also moved an unused import away from `utils.py`, where it's not used, to the two places it *is* used - `test_utils` and `generate-paas-manifest.py`

[Ticket](https://trello.com/c/H2ubPYqz/306-enable-autoescape-on-jinja-templates)

### DIFF
--- a/dmaws/utils.py
+++ b/dmaws/utils.py
@@ -82,8 +82,8 @@ def template_string(string, variables, templates_path=None):
         variables: a dict of variables that the template can use
 
     Raises:
-        TemplateSyntaxError: there is an issue with the template syntax.
-        UndefinedError: the template could not find the value for a variable.
+        jinja2.TemplateSyntaxError: there is an issue with the template syntax.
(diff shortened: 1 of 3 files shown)


lfdebrux:
Do we need a leave a comment to the same effect as the commit message for whoever may happen to read it?

dmaws/utils.py
```diff
@@ -3,7 +3,6 @@
 
 import yaml
 import jinja2
-from jinja2.exceptions import TemplateSyntaxError, UndefinedError  # noqa
```
katstevens: Hmm, we seem to have left these in for a reason - another file imports them

jonodrew: I couldn't find where else imports TemplateSyntaxError. As for UndefinedError, I've just added that specific import into test_utils rather than importing it to here and then importing it from here to there

PR #615, opened 2019-11-11
increase antivirus timeout at startup from 60 to 90 seconds

agalamatis:
trello: https://trello.com/c/JDKeSdiu/1245-increase-antivirus-apis-manifest-startup-timeout-setting

The timeout is also limited by the could controller's maximum_health_check_timeout which is 180 by default.

### DIFF
--- a/paas/antivirus-api.j2
+++ b/paas/antivirus-api.j2
@@ -1,5 +1,9 @@
 {% extends "_base.j2" %}
 
+{% block extra %}
+    timeout: 90
+{% endblock %}
+
 {% block env %}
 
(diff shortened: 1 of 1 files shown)


lfdebrux:
> The timeout is also limited by the could controller's maximum_health_check_timeout which is 180 by default.

Ooh that's interesting, where is that documented?

agalamatis:
> > The timeout is also limited by the could controller's maximum_health_check_timeout which is 180 by default.
> 
> Ooh that's interesting, where is that documented?

https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#timeout

PR #619, opened 2019-11-26
Enable Travis checks for Python 3.7 and 3.8

katstevens:
https://trello.com/c/GsDBHBo7/135-switch-on-python-37-and-38-travis-checks-to-check-potential-dependency-problems-before-upgrade

- No dependency bumps so far 🤞

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
   - "3.6"
+  - "3.7"
+  - "3.8"
 env:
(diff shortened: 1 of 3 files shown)


risicle:
I do get `boto3 1.10.25 has requirement botocore<1.14.0,>=1.13.25, but you'll have botocore 1.13.8 which is incompatible.` with this `requirements.txt` but it's not like this is the only place with that problem.

katstevens:
@risicle Yep, same for me - not too concerned.

PR #620, opened 2019-11-29
Add role to user name and email address in cleaned db dumps

lfdebrux:
This is a straw-person PR to go with [this tech debt ticket](https://trello.com/c/Ry6E8Xzk/1152-change-the-database-cleaning-script-to-make-user-email-addresses-more-descriptive).

### DIFF
--- a/scripts/clean-db-dump.sql
+++ b/scripts/clean-db-dump.sql
@@ -2,8 +2,8 @@
 
 \echo 'Update users'
 UPDATE users
-SET name = 'Test user',
-    email_address = id || '@user.marketplace.team',
+SET name = 'Test ' || role,
+    email_address = id || '@' || role || '.marketplace.team',
     password = :bcrypt_password,
     failed_login_count = 0,
     phone_number = '01234567890';


scripts/clean-db-dump.sql
```diff
@@ -2,8 +2,8 @@
 
 \echo 'Update users'
 UPDATE users
-SET name = 'Test user',
-    email_address = id || '@user.marketplace.team',
+SET name = 'Test ' || role,
+    email_address = id || '@' || role || '.marketplace.team',
```
katstevens: This will make a few things fail - `user.marketplace.team` is a safelisted domain for both buyers and admins. I believe we've also set up Notify to only pretend to send emails to that domain.

lfdebrux: Yeah, I had a feeling there would be issues like this; do you think the change is worth the hassle?

katstevens: Let's go back to the user need. When/where do we need to know what role the user has?

lfdebrux: Good question.

This PR was originally prompted by this [Slack thread comment by @aliuk2012]( https://gds.slack.com/archives/CFLL8967N/p1575015824006500?thread_ts=1575015193.005000&cid=CFLL8967N), as Al said there this change probably wouldn't help him, as his need is being able to see is logged into the system just by looking at the screens.

However it does give more context when you are on the account page of a user; and if the user name is on the page for some reason (say for instance in the header in some way) it would be informative.

lfdebrux:
This ticket has been moved to Won't Do

PR #622, opened 2019-12-06
168-spike-check-a-newly-deployed-app-can-connect-to-its-paas-backing-service-timebox-1d

benvand:
Make APIs use backing service checking healthcheck to ensure apps can connect to services
    
https://docs.google.com/document/d/18JV_ZTUqM500WLBxOt0lKtV12IYcS6Ay1SwJmxDOYb0
https://trello.com/c/ZxGiuNQb/1092-spike-check-a-newly-deployed-app-can-connect-to-its-paas-backing-service-timebox-1d

### DIFF
--- a/paas/api.j2
+++ b/paas/api.j2
@@ -1,5 +1,10 @@
 {% extends "_base.j2" %}
 
+{% block health_check %}
+    health-check-type: http
+    health-check-http-endpoint: {{ routes[0].partition('/')[1:]|join("") }}/_status
+{% endblock %}
+
 {% block env %}
       AWS_ACCESS_KEY_ID: {{ aws_access_key_id }}
       AWS_SECRET_ACCESS_KEY: {{ aws_secret_access_key }}
--- a/paas/search-api.j2
+++ b/paas/search-api.j2
@@ -1,5 +1,10 @@
 {% extends "_base.j2" %}
 
+{% block health_check %}
+    health-check-type: http
+    health-check-http-endpoint: {{ routes[0].partition('/')[1:]|join("") }}/_status
+{% endblock %}
+
 {% block env %}
       AWS_ACCESS_KEY_ID: {{ aws_access_key_id }}
       AWS_SECRET_ACCESS_KEY: {{ aws_secret_access_key }}


paas/api.j2
```diff
@@ -1,5 +1,10 @@
 {% extends "_base.j2" %}
 
+{% block health_check %}
+    health-check-type: http
+    health-check-http-endpoint: {{ routes[0].partition('/')[1:]|join("") }}/_status
```
katstevens: Does this depend on a fixed ordering of routes? I remember that biting us in the past...

lfdebrux:
This change makes sense to me, I'd like to know though before merging whether it's going to impact the service much

risicle:
I see problems when this is used against `-api` - it has the search api as one of its dependencies and its status endpoint considers it to be one too:

https://github.com/alphagov/digitalmarketplace-api/blob/fa6866bed73793679af8a3d9fdac9f5639f53801/app/status/views.py#L25

Passing the `search_api_client` to `get_app_status` causes it to consider it as a dependency. This would mean that we wouldn't be able to deploy a `-api` if the `-search-api` was down. Also if `-search-api` went down, all our `-api`s would get killed because their healthchecks would fail.

benvand:
@risicle Good call
What about something like this?
https://github.com/alphagov/digitalmarketplace-antivirus-api/pull/59
https://github.com/alphagov/digitalmarketplace-utils/pull/550

paas/api.j2
benvand: It does. It uses the same pattern as where it inherited it before here:
https://github.com/alphagov/digitalmarketplace-aws/blob/master/paas/_base.j2#L15

We could probably hard code this now we know what app we're operating on.

But might not need to change this at app now

alphagov/digitalmarketplace-antivirus-api#59
alphagov/digitalmarketplace-utils#550

risicle:
Yeah I think I added "additional checks" for that reason. Wanted to consider clamd an "internal" dependency.

benvand:
@katstevens @lfdebrux @risicle 

Thanks for the comments all, closing this in favour of a new approach changing what is called by `ignore-dependencies` by changing the underlying `get_app_status` in 

alphagov/digitalmarketplace-antivirus-api#59
alphagov/digitalmarketplace-utils#550

This will only affect the healthchecks (pingdom cannot hit apis). It means that `additional_checks` will always be called by the healthcheck request but defines a new arg called `additional_checks_extended` which will not run if `ignore-dependencies` is passed. See CHANGELOG.md in utils for more info.

PR #674, opened 2020-06-16
Increase time window for all alerts to 5 minutes

phantas:
~~**DO NOT MERGE OR APPLY BEFORE #675**~~ READY TO MERGE

Trello: https://trello.com/c/fF0V5auU

Please check commit messages for more information.

Note there are two actions that should follow after this PR has been merged:

- [x]  Revisit the notifications when resetting passwords for admin manager and if it's fit for the intended purpose. This will be a tech debt card
- [ ] Test that the change to the missing logs alarm keeps the alert of downtime at ~15 minute mark. To be tested before release to production

### DIFF
--- a/terraform/accounts/main/alarms.tf
+++ b/terraform/accounts/main/alarms.tf
@@ -22,9 +22,9 @@ resource "aws_cloudwatch_metric_alarm" "jenkins_data_volume_disk_space" {
     path = "/data"
   }
 
-  // For for every 60 seconds
+  // For for every 300 seconds
   evaluation_periods = "1"
-  period             = "60"
+  period             = "300"
 
   // If totals 10gb or lower (in bytes)
   statistic           = "Sum"
@@ -48,9 +48,9 @@ resource "aws_cloudwatch_metric_alarm" "jenkins_main_volume_disk_space" {
     path = "/"
   }
 
-  // For for every 60 seconds
+  // For for every 300 seconds
   evaluation_periods = "1"
-  period             = "60"
+  period             = "300"
 
   // If totals 1gb or lower (in bytes)
   statistic           = "Sum"
--- a/terraform/accounts/main/alarms.tf
+++ b/terraform/accounts/main/alarms.tf
@@ -22,7 +22,7 @@ resource "aws_cloudwatch_metric_alarm" "jenkins_data_volume_disk_space" {
     path = "/data"
   }
 
-  // For for every 300 seconds
+  // For every 300 seconds
   evaluation_periods = "1"
   period             = "300"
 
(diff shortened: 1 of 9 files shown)


terraform/modules/logging/alarms/missing-logs/alarms.tf
```diff
@@ -16,10 +16,9 @@ resource "aws_cloudwatch_metric_alarm" "missing_logs_alarm" {
   }
 
   // For 5 minutes
-  // This appears to cause our alarms to trigger at around the 10-12 minute mark
-  evaluation_periods = "5"
 
-  period = "60"
+  evaluation_periods = "1"
+  period = "300"
```
katstevens, line 21: I think this is the line that's failing the `terraformat` linting.

terraform/accounts/main/alarms.tf
```diff
@@ -22,9 +22,9 @@ resource "aws_cloudwatch_metric_alarm" "jenkins_data_volume_disk_space" {
     path = "/data"
   }
 
-  // For for every 60 seconds
+  // For for every 300 seconds
```
katstevens, line 25: I know these were already there, but can we get rid of the extra `for` in the comments 😸

terraform/modules/logging/alarms/admin-manager-password-reset/alarms.tf
```diff
@@ -6,9 +6,9 @@ resource "aws_cloudwatch_metric_alarm" "admin_manager_password_reset" {
   namespace   = "DM-admin-manager-password-reset"
   metric_name = "${var.environment}-admin-manager-password-reset"
 
-  // For for every 60 seconds
+  // For for every 300 seconds
   evaluation_periods = "1"
-  period             = "60"
+  period             = "300"
```
katstevens, line 11: I agree with the commit message - as discussed elsewhere, if the attacker is sophisticated enough to piggyback on a valid reset, 1 minute vs 5 minutes is unlikely to make a difference. However we should make a note to review this as a key alert if/when we get set up with Splunk.

terraform/modules/logging/alarms/missing-logs/alarms.tf
phantas: Ack! My bad!, completely forgot to run terraform plan...

terraform/accounts/main/alarms.tf
phantas: I did think about this, actually. And wondered if it was meant as `for "for every 300 seconds"` 🤣 I will clean up

terraform/modules/logging/alarms/admin-manager-password-reset/alarms.tf
phantas: I'll add the comment about Splunk to the tech debt card

phantas: tech debt card added, resolving

phantas:
When running plan, I noticed the node version for the lambda live on Preview did not match the one on our code. This is fixed and upgraded to the latest supported one on #675 which should be merged and applied before this one

PR #675, opened 2020-06-17
Upgrade terraform

phantas:
See Trello: https://trello.com/c/X4v1uBUh

Just maintenance stuff, ensuring we are using the latest and it's consistent across. 

I ran `terraform plan` against Preview and all is :+1:

### DIFF
--- a/terraform/modules/log-streaming/lambda.tf
+++ b/terraform/modules/log-streaming/lambda.tf
@@ -59,7 +59,7 @@ resource "aws_lambda_function" "log_stream_lambda" {
   source_code_hash = "${data.archive_file.lambda_zip.output_base64sha256}"
   role             = "${aws_iam_role.lambda.arn}"
   handler          = "main.handler"
-  runtime          = "nodejs8.10"
+  runtime          = "nodejs10.x"
   memory_size      = 128
   timeout          = 10
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "3.7"
   - "3.8"
 env:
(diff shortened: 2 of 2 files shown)


terraform/modules/log-streaming/lambda.tf
```diff
@@ -59,7 +59,7 @@ resource "aws_lambda_function" "log_stream_lambda" {
   source_code_hash = "${data.archive_file.lambda_zip.output_base64sha256}"
   role             = "${aws_iam_role.lambda.arn}"
   handler          = "main.handler"
-  runtime          = "nodejs8.10"
+  runtime          = "nodejs12.x"
```
katstevens, line 62: Despite what it says here (😿), our preview/staging/prod lambdas are all currently on 10.x. Do we want to upgrade to 12?

phantas: That's a good question. I chose to bump into 12 because it's the current active LTS. I had a look at https://nodejs.org/tr/blog/uncategorized/10-lts-to-12-lts/ and the code and nothing raised any alarms. Do you think it's maybe best to scope out the move to 12 out of this PR to unblock the other PR?

phantas: On retrospect, having a look again, the code is quite old and I can't find the original code it's based off. I also don't want to impact the work @gidsg is doing, so I'll switch to 10.
We'll need to revisit in 6 months time when 10 reaches end of life, but I'll add a card to the backlog with a date to remind us.

phantas:
@katstevens changed to 10.x, ready for re-review

PR #680, opened 2020-06-23
Don't echo credentials (send to /dev/null)

gidsg:
https://trello.com/c/TNoXg0lb/147-2-postgres-secrets-logged-to-console-when-db-backup-job-fails

Previously in the error case the credentials were shown as they were printed to stdout in the PaaS logs. This still executes the command but forwards to `/dev/null` so nothing is printed.

We can confirm this is working on Jenkins as expected once merged.

### DIFF
--- a/db-backup/create-db-dump.sh
+++ b/db-backup/create-db-dump.sh
@@ -1,9 +1,7 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-set +x  # Don't log credentials
-DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
-set -x  # Restore logging
+DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri') >/dev/null # Don't print credentials to stdout
 echo -n "${PUBKEY}" > /app/public.key
 gpg2 --import /app/public.key
 pg_dump "${DB_URI}" --no-acl --no-owner --clean --if-exists | gzip | \


lfdebrux:
> We can confirm this is working on Jenkins as expected once merged.

Usually with Jenkins PRs I find it's better to push the changes before merging, saves you having to open a hundred PRs to fix the same issue 😅

db-backup/create-db-dump.sh
```diff
@@ -1,9 +1,7 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-set +x  # Don't log credentials
-DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
-set -x  # Restore logging
+DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri') >/dev/null # Don't print credentials to stdout
```
lfdebrux, line 4: I'm not sure I understand why the `> /dev/null` is needed; if I do `TEST=$(echo hello world)` on my machine nothing gets echoed to stdout, so why is stuff getting echoed here? 🤔

gidsg: I think it shows up as part of `stderr` in the error case i.e. here:  https://ci.marketplace.team/job/database-backup/1137/console

gidsg: the `ERR` prefix maps to `stderr` as per https://docs.cloudfoundry.org/devguide/deploy-apps/streaming-logs.html

lfdebrux: So should we be redirecting `stderr` instead of `stdout`?

gidsg: Possibly, that said I'm not sure if that'd work either thinking more about it. If it is just dumping the command history.

gidsg:
Closing this whilst I think some more about it and talking to PaaS support.

PR #681, opened 2020-06-23
do not expand console output

gidsg:
https://trello.com/c/TNoXg0lb/147-2-postgres-secrets-logged-to-console-when-db-backup-job-fails

The effect of the `set -x` argument was to print and expand all arguments which caused the printing of secrets into the logs.

Please documentation at: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

This also switches to use the PaaS built-in [DATABASE_URL](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#DATABASE-URL) rather than parsing VCAP using jq.

You can see the output of this branch for a failure case here: https://ci.marketplace.team/job/database-backup/1166/execution/node/161/log/

### DIFF
--- a/db-backup/create-db-dump.sh
+++ b/db-backup/create-db-dump.sh
@@ -1,12 +1,9 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-set +x  # Don't log credentials
-DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
-set -x  # Restore logging
 echo -n "${PUBKEY}" > /app/public.key
 gpg2 --import /app/public.key
-pg_dump "${DB_URI}" --no-acl --no-owner --clean --if-exists | gzip | \
+pg_dump "${DATABASE_URL}" --no-acl --no-owner --clean --if-exists | gzip | \
   gpg2 --trust-model always -r "${RECIPIENT}" --out /app/"${DUMP_FILE_NAME}" --encrypt
 
 /usr/local/bin/python /app/upload-dump-to-s3.py
--- a/db-backup/create-db-dump.sh
+++ b/db-backup/create-db-dump.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euxo pipefail
+set -euo pipefail
 
 echo -n "${PUBKEY}" > /app/public.key
 gpg2 --import /app/public.key


phantas:
Was there a reason why we had the -x to start with? Are we loosing anything by being less verbose?
I had a look at the failed and successful jobs in production and the recent one in preview and I couldn't tell.

db-backup/create-db-dump.sh
```diff
@@ -1,12 +1,9 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-set +x  # Don't log credentials
-DB_URI=$(echo $VCAP_SERVICES | jq -r '.postgres[0].credentials.uri')
-set -x  # Restore logging
 echo -n "${PUBKEY}" > /app/public.key
 gpg2 --import /app/public.key
-pg_dump "${DB_URI}" --no-acl --no-owner --clean --if-exists | gzip | \
+pg_dump "${DATABASE_URL}" --no-acl --no-owner --clean --if-exists | gzip | \
```
phantas, line 6: 😍 this is a great catch!

gidsg:
> Was there a reason why we had the -x to start with? 

@phantas Good question, I am not 100% sure to be honest as this was before my time. It does produce verbose log output for debugging i.e. you can see all commands and their results sent to stderr (ERR) e.g. in https://ci.marketplace.team/job/database-backup/1137/consoleText

phantas:
> > Was there a reason why we had the -x to start with?
> 
> @phantas Good question, I am not 100% sure to be honest as this was before my time. It does produce verbose log output for debugging i.e. you can see all commands and their results sent to stderr (ERR) e.g. in https://ci.marketplace.team/job/database-backup/1137/consoleText

Yeah, that was the job I was looking at, as well as 1138 (success). The `upload_to_s3.py` already does some reasonable printing on failures and the job only really captures the output when it fails.
I had a look at the original PR #315 and there's nothing specific to that flag, so I am happy to approve 👍

db-backup/create-db-dump.sh
katstevens: Nice find! Did you try a success run on Preview with this new environment variable?

gidsg: @katstevens whilst I forced the jobs to fail by constraining the memory, they did get as far as connecting to the database and starting following steps before running out of memory, so I am confident this works.

katstevens: Excellent stuff 👍

PR #687, opened 2020-07-08
add splunk service to apps

gidsg:
https://trello.com/c/yJeKtmkj/201-2-setup-splunk-on-our-apps-for-log-aggregation-timebox-2-days
add splunk service to apps. This is as-per https://github.com/alphagov/tech-ops/blob/master/cyber-security/components/csls-splunk-broker/docs/user-guide.md#bind-a-splunk-service-to-your-apps-using-an-app-manifest

### DIFF
--- a/vars/common.yml
+++ b/vars/common.yml
@@ -9,7 +9,8 @@ router:
     - antivirus-api.{env}.marketplace.team
     - assets.{env}.marketplace.team
   services:
-    - digitalmarketplace_prometheus
+    - digitalmarketplace_prometheus 
+    - digitalmarketplace_splunk
 
 api:
   routes:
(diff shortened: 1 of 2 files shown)


lfdebrux:
I think this is going to add splunk as a service to all apps in all spaces? Is that what we want right now?

gidsg:
@lfdebrux I think so. My plan was to re-release all apps in Preview first so I can see it is working as expected before rolling out further. Although there is a small risk with this strategy that it may be inadvertently put into production by one of the apps deploying.

lfdebrux:
> @lfdebrux I think so. My plan was to re-release all apps in Preview first so I can see it is working as expected before rolling out further. Although there is a small risk with this strategy that it may be inadvertently put into production by one of the apps deploying.

I think we might need some time testing Splunk in preview before we decide whether/how we want to use it in all spaces, so we should probably keep it just to preview for now if that's alright.

gidsg:
> we should probably keep it just to preview 

@lfdebrux OK I have made this change in 9c4d5b0 and tested using `make generate-manifest`. 
Note as this adds some complexity to these files, we should aim to revert this and move to common as soon as we are comfortable rolling out Splunk more widely.

PR #704, opened 2020-07-30
increase supplier frontend instances and memory

gidsg:
scaling up in anticipation of increased traffic


katstevens:
Can we bump the `api` as well (to 20)?

PR #707, opened 2020-08-03
Revert "increase supplier frontend instances and memory"

gidsg:
Reverts alphagov/digitalmarketplace-aws#704

scaling back down after anticipated increase in traffic.


katstevens:
@gidsg I think we can merge this now?

gidsg:
@katstevens thanks for the nudge, can you approve?

PR #733, opened 2020-09-24
Make it clear we're using terraform 0.11.x

bjgill:
I've got terraform 0.13 installed, and was confused as to why it wasn't working.


lfdebrux (approved):
Good call

gidsg (approved):

PR #746, opened 2020-10-15
Fix bug in deploy-app recipe

lfdebrux:
https://trello.com/c/Kop14JBa/1425-spike-our-zero-downtime-deploy-script-can-fail-in-a-way-that-requires-manual-recovery

Allow the commands to delete the previous app in the deploy-app recipe to fail [[1]]. If for some reason the existing app has the wrong name or isn't present, we still want the deployed app to be renamed to have the right name.

[1]: https://www.gnu.org/software/make/manual/html_node/Errors.html#Errors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,8 +122,8 @@ deploy-app: ## Deploys the app to PaaS
 	@echo "Waiting for previous app version to process existing requests..."
 	sleep 60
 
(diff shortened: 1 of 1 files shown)


bjgill (approved):
:tada:

PR #750, opened 2020-10-20
Upgrade to terraform 0.12

gidsg:
https://trello.com/c/usuc7ANJ/79-1-upgrade-aws-repo-to-use-terraform-v012-timebox-05-day
see https://www.terraform.io/upgrade-guides/0-12.html

Although 0.13 has been released this upgrades to use 0.12 as the recommendation is to upgrade only one major version at a time.

Depends on https://github.com/alphagov/digitalmarketplace-credentials/pull/368 and https://github.com/alphagov/digitalmarketplace-credentials/pull/370

## Methodology
* Ran `terraform init` on every module and resolved any errors
* Ran `terraform 0.12upgrade` on every module, resolving any manual TODOs.
* Ran `make refresh` for every account and fixed errors and warnings.
* Ran `make refresh` for every environment and fixed errors and warnings.

## Notes
* Standardised on a single AWS provider version `2.70` as this was causing many errors either with no version or an older version being declared in some modules
* Upgraded `csw-client-role` module to a version compatible with 0.12
* I did not run `apply` on every module before doing this but this should be fine as long as we don't have any un-applied state committed in the repo 🤞


bjgill (approved):
I skimmed the diff and nothing looked wrong. Given that you made the changes programmatically,  I'm happy they're likely correct.

What's the reason for not running `apply` for each module? Does it just take too long? I guess I'm just a bit nervous about only discovering a problem at some future point when we try to apply another change...

gidsg:
@bjgill yes a combination of it would take a long time plus the way we wrap the commands in the `Makefile` I was not confident doing a vanilla Terraform apply would have the same outcome.

gidsg:
Closing this whilst I investigate an issue with credentials.

bjgill:
Re-opening and converting to a draft so I can continue working on this.

lfdebrux:
Can we add a [`.terraform-version` file](https://github.com/tfutils/tfenv#terraform-version-file) somewhere in this repo so we know the exact version of terraform to use?

lfdebrux (changes requested):
I can't `make plan` in `environments/preview`, I get a bunch of `Invalid character` errors

terraform/modules/router/route53.tf
```diff
@@ -1,118 +1,124 @@
...
 resource "aws_route53_zone" "root" {
-  name = "${var.domain}"
+  name = var.domain
 }
 
 resource "aws_route53_record" "www_acm_validation" {
-  zone_id = "${aws_route53_zone.root.zone_id}"
-  name    = "${var.www_acm_name}${var.domain}"
+  zone_id = aws_route53_zone.root.zone_id
+  name    = var.www_acm_name}${var.domain
```
lfdebrux, line 12: This looks a little suspect...

bjgill: Yes - I think this may be the cause of the error you were seeing for Preview. We'll take a look.

terraform/modules/logging/alarms/missing-logs/alarms.tf
```diff
@@ -3,16 +3,16 @@
 // The alarm will be triggered when the metric is not emitted due to the breaching poicy on missing data
 
 resource "aws_cloudwatch_metric_alarm" "missing_logs_alarm" {
-  count             = "${length(var.app_names)}"
-  alarm_name        = "${var.environment}-missing-logs-${var.app_names[count.index]}"
+  count             = length(var.app_names)
+  alarm_name        = var.environment}-missing-logs-${var.app_names[count.index]
```
lfdebrux, line 7: Some extra curly braces snuck in here

bjgill: Oh dear. I'd recommend not reviewing further - probably more efficient for us to fix all of these first.

bjgill:
Right - we'd forgotten about the environments, which is how we missed all those errors. We've fixed the errors, and this now works for all environments and accounts.

bjgill:
I've added CI on top of this branch in another PR, so I'm fairly certain that it is now syntactically valid

https://github.com/alphagov/digitalmarketplace-aws/pull/834

lfdebrux (approved):

lfdebrux (approved):
Works for me! Nice one ⛰️

PR #751, opened 2020-10-22
add alex smith IAM user

gidsg:
https://trello.com/c/XTBQpMBa/1986-add-alex-to-some-of-the-things
add @alex9smith IAM user.


bjgill (approved):
:tada: I assume this is non-prod only?

gidsg:
@bjgill this does not grant any permissions, only creates the IAM user I believe. The permissions are done in a separate step via credentials repo.

bjgill:
Ah. I see. :+1:

PR #754, opened 2020-10-29
Use Docker credentials

gidsg:
https://trello.com/c/TM0mwfiE/2014-add-dockerhub-credentials-to-paas-manifests

Note `CF_DOCKER_PASSWORD` should be automatically picked up and used as long as the env var is set. See https://docs.cloudfoundry.org/devguide/deploy-apps/push-docker.html#private-repo


lfdebrux (approved):
Let's take it for a spin!

PR #758, opened 2020-11-03
Scale up staging for load testing

bjgill:
Trello: https://trello.com/c/kGmsXWZ1/520-3-find-out-whether-dmp-can-cope-with-dos5-closing

We will shortly be running load testing on staging. The goal is to find out whether the DMp is likely to be able to handle a reasonable worst-case scenario for traffic during DOS 5 closing.

For this test to give us meaningful results, we need staging to match production as it will be on the day of DOS 5 closing. Thus, scale up staging. I've modelled the scale after that we used for DOS 5 opening and G12 closing.

Also disable rate-limiting, since we are running the load test from a single IP.

### DIFF
--- a/vars/staging.yml
+++ b/vars/staging.yml
@@ -4,8 +4,8 @@ maintenance_mode: live
 instances: 5
 
 api:
-  memory: 2GB
-  instances: 10
+  memory: 4GB
+  instances: 20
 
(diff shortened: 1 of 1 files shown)


gidsg (approved):
LGTM 👍 🆙

PR #769, opened 2020-11-12
enable redis env var on preview only

gidsg:
https://trello.com/c/0XO8CHaP/381-05-stand-up-the-paas-infrastructure-for-the-preview-environment

We want to test redis initially on preview only. This is behind an env var. To set this on preview only this also introduces some logic to allow per-environment additions to the env block of the manifest.

I have tested this using `make generate-manifest`

### DIFF
--- a/paas/_base.j2
+++ b/paas/_base.j2
@@ -34,6 +34,11 @@ applications:
 
       DM_LOG_PATH: ''
 
+      {% if env|default([])|length > 0 %}
+        {% for k, v in env.items() %}
+      {{ k }}: {{ v }}
+        {% endfor %}
+      {% endif %}
       {% block env %}
       {% endblock %}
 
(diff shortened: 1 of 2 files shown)


alex9smith (commented):
I built the manifest for buyer-frontend on preview and noticed that `DM_USE_REDIS_SESSION_TYPE` was set to `True` in the output, rather than `true` which is what you wrote in the vars YAML. Is that going to cause any problems?

gidsg:
@alex9smith in this case, it should not cause any issues as we are only looking for the presence of the env var not its value but I can put the value in quotes to be more explicit.

alex9smith:
@gidsg No, that's fine and what I thought as well - just wanted to double check. Let's leave it as is since this env var is only going to exist for a short amount of time while we rollout this change anyway.

alex9smith (approved):

PR #792, opened 2020-11-24
add redis env var and service to common vars

gidsg:
https://trello.com/c/WXJhhH1z/571-1-deploy-the-new-user-session-management-to-production

Add Redis service and env var to common config as we are enabling it on all environments.


bjgill (approved):
Looks reasonable. Presumably we can remove `DM_USE_REDIS_SESSION_TYPE` once this is live in all environments?

gidsg:
> Presumably we can remove DM_USE_REDIS_SESSION_TYPE once this is live in all environments?

@bjgill that is the eventual goal but as this check is in utils we will have to roll out changes to all apps to do that .

lfdebrux:
Tested with `make generate-manifest APPLICATION_NAME=buyer-frontend STAGE=production`, all looks to be in order 👍

PR #796, opened 2020-11-27
Change Jenkins disk space alarms to use Minimum

alex9smith:
This was changed in the AWS console by @bjgill as part of
https://trello.com/c/tlhoS1vi/2088-disk-space-alarms-for-jenkins-are-reporting-incorrect-values

Update the config in Terraform as well so we don't revert the changes when making future infrastructure changes. I added the `datapoints_to_alarm` variable because Terraform was complaining without it and we do want to alarm when we get 1 datapoint:

```
Error: aws_cloudwatch_metric_alarm.jenkins_data_volume_disk_space: expected datapoints_to_alarm to be at least (1), got 0
```


bjgill (approved):
Good catch!

PR #797, opened 2020-11-27
Pass list of suppliers to apps

bjgill:
Trello: https://trello.com/c/KikQx3R2/

We will be using this configuration to show the G12 recovery process only to the suppliers involved.

Environment variables must be strings, so convert the list to a comma-separated string.

This relies on https://github.com/alphagov/digitalmarketplace-credentials/pull/359, which adds the necessary fields into the credentials.

### DIFF
--- a/paas/api.j2
+++ b/paas/api.j2
@@ -10,4 +10,6 @@
 
       DM_SEARCH_API_AUTH_TOKEN: {{ search_api.auth_tokens[0] }}
       DM_SEARCH_API_URL: https://dm-search-api-{{ environment }}.cloudapps.digital
+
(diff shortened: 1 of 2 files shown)


bjgill:
@agalamatis pointed out that the API app will also need access to the list.

lfdebrux (approved):

bjgill:
We discussed via Slack with Fajer.

It's OK for the existence of the recovery process to be public - especially as to try to hide it would be a lot of work.

However, we don't want to make the list of suppliers involved public, which is why we're putting them encrypted into the credentials.

PR #823, opened 2021-01-21
Update postgresql in `make run-postgres-container` from 9.5 to 10.13

lfdebrux:
We are now using PostgreSQL 10.13 in production, we should make sure our database backup pipeline uses the same or newer version.

I have verfied that the Docker images involved exist and can be built, once this PR is approved I will push `digitalmarketplace/db-backup` to Docker Hub.


gidsg (approved):
Looks good to me 🍡 👍

PR #827, opened 2021-01-27
bump postgres container version

gidsg:
https://trello.com/c/XjcK0K3g/737-upgrade-environments-to-postgres-12
Upgrade postgres container to match what we are running on PaaS﻿


lfdebrux (approved):

lfdebrux:
We might also need to figure out a way to bump the postgres-cli version in `db-dump/Dockerfile`, as I think the latest version in the Debian repos is version 11.

PR #828, opened 2021-01-27
Pass list of draft services to supplier-frontend

bjgill:
Trello: https://trello.com/c/13Ryt5o5/679-5-as-a-supplier-i-cannot-see-submit-any-draft-services-that-ccs-will-not-consider

We will be using this configuration to hide draft services not involved in the G12 recovery process.

Environment variables must be strings, so convert the list to a comma-separated string.

This relies on https://github.com/alphagov/digitalmarketplace-credentials/pull/366, which adds the necessary fields into the credentials.


alex9smith (approved):
Looks sensible to me!

PR #834, opened 2021-02-03
Add terraform testing to CI

bjgill:
Inspired by [a recent PR](https://github.com/alphagov/digitalmarketplace-aws/pull/750), where we wasted a lot of time on manual testing.

Add a script that we run in CI. This will check that our terraform code is syntactically correct, internally consistent, and properly formatted. It does not try to access external state, so is safe to run in CI.

Needs Terraform 0.12, so this PR depends on https://github.com/alphagov/digitalmarketplace-aws/pull/750


lfdebrux (approved):
💯

PR #840, opened 2021-02-12
Upgrade backup task for postgres 12

gidsg:
https://trello.com/c/mZ1ZDfS1/791-3-db-backup-app-is-running-end-of-life-os

Updates container and client to use postgres 12. Have tested successfully against preview: https://ci.marketplace.team/job/database-backup/1393/

## Notes
* `postgresql-client-12` is not available in the default package repository, therefore this adds the postgres repository as per the guidance in: https://www.postgresql.org/download/linux/ubuntu/
* This should not be merged until Postgres 12 has been rolled out to all environments 
* I will push the docker image once this PR is merged

### DIFF
--- a/db-backup/Dockerfile
+++ b/db-backup/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-slim
+FROM python:3.6.12-slim
 
 RUN mkdir /app
 
--- a/db-backup/Dockerfile
+++ b/db-backup/Dockerfile
@@ -4,7 +4,19 @@ RUN mkdir /app
 
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y --no-install-recommends gnupg2 jq wget vim curl postgresql-client
+    apt-get install -y --no-install-recommends gnupg2 jq wget vim curl 
+
+# Create the file repository configuration:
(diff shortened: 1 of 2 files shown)


lfdebrux:
> This should not be merged until Postgres 12 has been rolled out to all environments

I think the client should be backwards compatible no? https://www.postgresql.org/docs/12/app-pgdump.html#PG-DUMP-NOTES

gidsg:
@lfdebrux yes, the client should be backwards compatible but I believe the Docker image i.e. that specified in https://github.com/alphagov/digitalmarketplace-aws/pull/840/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R191 must match the running version.

db-backup/Dockerfile
```diff
@@ -1,10 +1,22 @@
-FROM python:3.6-slim
+FROM python:3.6.12-slim
```
alex9smith, line 1: Any reason not to go to `python:3.6.13-slim` now?

alex9smith (approved):
Looks sensible 👍  Just one question

db-backup/Dockerfile
gidsg: 👀  good spot @alex9smith updated now in 79cd77a

PR #860, opened 2021-03-10
Update CodeCommit repository default branches

alex9smith:
We've started to change our default git branches from `master` to `main`. Update our Terraform configuration to be able to handle different default branch names, and update the branch names for the repos we've already changed.

Also switch to the new (in 0.12.6) `for_each` Terraform operator which makes things a bit easier.

Applying these changes will destroy and re-create all of our CodeCommit resources because they're currently saved in the Terraform state by list index, and this PR changes each repo to a map key. I have checked the Jenkins job which does the backup and I don't think this will cause problems, but I'll run the job after applying to make sure.

I'll also add an item to the 'Tasks per-repo' check list on the attached Trello card to remind us to also update this file with the new default branch name.

https://trello.com/c/DrVWMkcZ/2169-rename-default-branches-on-our-repos-from-master-to-main

### DIFF
--- a/terraform/accounts/main/codecommit.tf
+++ b/terraform/accounts/main/codecommit.tf
@@ -1,45 +1,107 @@
 variable "alphagov_git_repositories" {
-  type = list(string)
+  type = map(object({
+    default_branch = string
+  }))
 
-  default = [
-    "digitalmarketplace-api",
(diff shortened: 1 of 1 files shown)


bjgill (approved):
Neat.

PR #874, opened 2021-03-30
Upgrade Terraform AWS provider to v3.34.0

alex9smith:
Apologies in advance, this is a sizeable PR but it's not easy to split up.

In order to use the Node 14 Lambda runtime instead of Node 10 for our log streaming Lambda[[1]], we also need to upgrade the AWS provider.

Some of our modules had provider versions defined and some didn't, so I've taken the opportunity to clean things up and only define provider versions in the places we can run Terraform from - `main.tf` in each account and environment folder. The modules will get the correct version of the provider passed down to them when they're loaded.

We are on Terraform v.0.12, but plan to upgrade to 0.13 and 0.14 soon. Use the new `required_providers` syntax for the version definitions so this change is compatible with both Terraform v0.12 and v0.13[[2]].

Also set a `AWS_DEFAULT_REGION` environment variable for validation to work around a limitation in `terraform validate` in 0.12[[3]]

Applying these changes will recreate the Jenkins DNS records and change the ARN for our Cloud Watch group from `"arn:aws:logs:eu-west-1:398263320410:log-group:digitalmarketplace-main-account-cloudtrail-cloudtrail-log-group:*"` to `"arn:aws:logs:eu-west-1:398263320410:log-group:digitalmarketplace-main-account-cloudtrail-cloudtrail-log-group"` (removing `:*` from the end). ARNs are generated by AWS so this appears to be a change coming from updating the provider version. It's also updated in all the places that use this ARN (eg. the access policies), and I don't think dropping the `:*` changes the meaning of anything we do with this ARN. Therefore, I don't expect it to cause any problems. I'll apply changes to the development account and preview environment first and check Cloud Watch is still working to make sure though.

https://trello.com/c/sBqxdTZC/1734-upgrade-logstream-lambda-to-latest-active-lts

[1]: https://github.com/alphagov/digitalmarketplace-aws/blob/master/terraform/modules/log-streaming/lambda.tf#L73
[2]: https://www.terraform.io/docs/language/providers/requirements.html#v0-12-compatible-provider-requirements
[3]: https://github.com/hashicorp/terraform/issues/21408

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ terraformatest:
 
 .PHONY: test-terraform
 test-terraform:
+	$(eval export AWS_DEFAULT_REGION=eu-west-2)
 	./scripts/validate-terraform.sh
 
 .PHONY: test-flake8
--- a/terraform/accounts/backups/main.tf
+++ b/terraform/accounts/backups/main.tf
@@ -1,8 +1,4 @@
 provider "aws" {
-  region = "eu-west-1"
-}
-
-provider "aws" {
   alias  = "london"
   region = "eu-west-2"
 }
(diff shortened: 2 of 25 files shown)


bjgill (approved):
Changes look good, and very well documented. Deployment plan seems sensible.

:rocket:

PR #876, opened 2021-03-31
Upgrade log streaming Lambda runtime from Node 10 to Node 14

alex9smith:
Node 10 is EOL on April 30th 2021 and will no longer receive security updates. Upgrade the runtime to Node 14, the most recent LTS version which is supported until April 2023.

I have had a quick scan of [the changelogs](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md) from Node 10 to 14 and I don't believe we are using anything that's been deprecated. I'll deploy to preview first and check logs are still arriving in Logit to make sure it's all still fine before doing the other environments.

https://trello.com/c/sBqxdTZC/1734-upgrade-logstream-lambda-to-latest-active-lts

### DIFF
--- a/terraform/modules/log-streaming/lambda.tf
+++ b/terraform/modules/log-streaming/lambda.tf
@@ -69,7 +69,7 @@ resource "aws_lambda_function" "log_stream_lambda" {
   source_code_hash = data.archive_file.lambda_zip.output_base64sha256
   role             = aws_iam_role.lambda.arn
   handler          = "main.handler"
-  runtime          = "nodejs10.x"
+  runtime          = "nodejs14.x"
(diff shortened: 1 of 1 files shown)


bjgill (approved):
Sounds good. Are you planning to check Splunk as well, or does that not rely on the λ?

alex9smith:
Splunk is different of course- application logs get read from stdout on each of the PaaS apps then forwarded on via a backing service `digitalmarketplace_splunk`. There's a bit of information in [the manual](https://alphagov.github.io/digitalmarketplace-manual/infrastructure/logging.html) about it.

PR #910, opened 2021-05-04
upgrade to Terraform 0.14

gidsg:
https://trello.com/c/zZEis2fo/860-3-upgrade-to-terraform-014

Note this version introduces a lock file which fixes plugin versions, it is recommended to check this into version control see: https://www.terraform.io/upgrade-guides/0-14.html#opting-out-of-dependency-locking

This also bumps the AWS provider version which seemed to be a useful way to fix a terraform upgrade error about legacy providers (although possibly just a side-effect of forcing a download of a new provider)


terraform/accounts/main/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.37.0"
```
bjgill: Why do we need to bump this version?

gidsg: @bjgill I got an error about legacy providers and the upgrade not being able to proceed before I did this. I think maybe the version bump makes it fetch the new provider version which solved the issue as a side-effect.

bjgill: Sounds reasonable. Would have been nice to have this information in the commit/PR description.

bjgill (approved):
:rocket:

phantas:
I was still reviewing but up till now I only had minor suggestions around commit messages so I'll just finish combing the files, that I was in the middle of, but, as far as I can see 👍 from me.

terraform/accounts/main/main.tf
gidsg: @bjgill fair point, I've updated the PR description now.

PR #911, opened 2021-05-05
Only allow encrypted connections to Terraform state buckets

alex9smith:
Define access policies for our Terraform state buckets that only permit encrypted connections. This addresses a medium-risk vulnerability from our recent ITHC.

These are the Terraform state buckets and therefore they cannot be created through Terraform or easily managed through a module. This is why the policies are defined in-line after the Terraform backend configuration in each AWS account.

https://trello.com/c/KeWS6zRV/925-644-bucket-allowing-clear-text-http-communication

### DIFF
--- a/terraform/accounts/backups/main.tf
+++ b/terraform/accounts/backups/main.tf
@@ -33,3 +33,23 @@ terraform {
   }
 }
 
+resource "aws_s3_bucket_policy" "backups_tfstate_bucket_policy" {
+  bucket = "digitalmarketplace-terraform-state-backups"
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Id = "BackupsTFStateOnlyEncryptedConnectionPolicy"
+    Statement = [
+      {
+        Effect: "Deny",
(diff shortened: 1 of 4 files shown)


bjgill (approved):
Sounds like a sensible approach to a slightly gnarly problem.

terraform/accounts/backups/main.tf
```diff
@@ -33,3 +33,23 @@ terraform {
   }
 }
 
+resource "aws_s3_bucket_policy" "backups_tfstate_bucket_policy" {
+  bucket = "digitalmarketplace-terraform-state-backups"
+  policy = jsonencode({
+    Version = "2012-10-17"
```
bjgill, line 39: Though, what's the significance of this date?

alex9smith: It means we're using the latest version of the policy language and we have to include it - see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

PR #927, opened 2021-05-25
Enable default S3 encryption

bjgill:
https://trello.com/c/HCf26VVT/922-3-645-bucket-without-default-encryption-enabled

Enable default encryption for all our S3 buckets. This was prompted by the recent IT health check. Security benefits likely to be minimal, but it's free and easy to enable, so why not.

We've tested that this shouldn't cause problems already - see 029e9ac. This commit adds the necessary config to all other buckets in all environments/accounts/etc.

Also terraformat picked up some unrelated formatting-only changes.


Krenair (approved):
Let's give it a go.

PR #936, opened 2021-06-08
Add assets URL to buyer frontend

bjgill:
https://trello.com/c/5E22SHaQ/988-ccs-sourcing-admins-cant-view-suppliers-modern-slavery-statements

It's currently hard-coded (https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/1e5b9e7def105d4c055b25c72fa7e6fd2acd49e6), which we want to remove.

### DIFF
--- a/paas/buyer-frontend.j2
+++ b/paas/buyer-frontend.j2
@@ -12,7 +12,8 @@
       DM_REDIS_SERVICE_NAME: 'digitalmarketplace_redis'
       DM_SEARCH_API_AUTH_TOKEN: {{ search_api.auth_tokens[0] }}
       DM_SEARCH_API_URL: https://dm-search-api-{{ environment }}.cloudapps.digital
-      
+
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
👍 seems like a useful thing to have available to the app. I suspect that if we just switch out the `.replace()` in [buyer frontend](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/1e5b9e7def105d4c055b25c72fa7e6fd2acd49e6#diff-1560286934bceea743ab16808319e4f863662d6581e7f9fbff0e24e8e48b8ee1R100) for this domain then we'll break all the document links on preview / staging because I don't think we populate the S3 buckets. That's a problem to solve in the next step though!

bjgill:
That's an excellent point that I hadn't considered. I'll merge this anyway because I think it'll still be useful in future. But I suspect it might be wise to hold off on switching the replace to this domain for the moment.

PR #941, opened 2021-06-14
Remove alarms that require no action

alex9smith:
We have a scraper which is triggering these alarms with 100's of requests every hour. We don't need to do anything when they're triggered so they're causing us unnecessary alarm fatigue and could result in us missing real problems. 

Increasing the trigger thresholds to over what the scraper causes will effectively cause the alarms to never trigger, which won't be any more use than simply deleting them. 

https://trello.com/c/sZqatZVp/4-2-reduce-alerts-that-require-no-action


gidsg (approved):
LGTM ⏰ 👍

PR #942, opened 2021-06-15
Enable MFA delete protection on our S3 buckets

alex9smith:
This was identified in our recent ITHC and we've already updated all the buckets in the AWS console. Update Terraform to match what's deployed.

If you checkout this branch and run `make plan` for one of preview, staging or production the only changes you should now see are removing some Cloudwatch alarms (see #941) and maybe a DNS validation update.


gidsg (approved):
Good spot 👀  👍 
we didn't remember to do this when initially doing the work via the console.

PR #945, opened 2021-06-15
use python 3.9

gidsg:
https://trello.com/c/PqMIVIGW/13-upgrade-apps-to-python-36-38

Similar to https://github.com/alphagov/digitalmarketplace-scripts/pull/679 we are upgrading python in our non-deployable apps first to reduce risk.


bjgill (approved):
Looks good. Am I right in thinking that this only ever gets run by humans, so we don't need to worry about Jenkins/etc.?

bjgill:
To merge this, it looks as if you'll need to remove some expected checks. I think you've got permissions to do so, but shout if not and I can.

gidsg:
> Am I right in thinking that this only ever gets run by humans, so we don't need to worry about Jenkins/etc.?

Good point on further investigation, it seems like some scripts in this repo are run by Jenkins, for example https://github.com/alphagov/digitalmarketplace-aws/blob/main/scripts/generate-paas-manifest.py so I have converted this PR to a draft for now.

gidsg:
Closing this as some of the scripts run on Jenkins which is still on Python 3.6

PR #956, opened 2021-07-16
Set up Ubuntu 20 LTS instance

gidsg:

This sets up a new Jenkins instance which is running Ubuntu 20 LTS which includes Python 3.8 in the package manager.

### DIFF
--- a/terraform/accounts/main/jenkins.tf
+++ b/terraform/accounts/main/jenkins.tf
@@ -22,6 +22,20 @@ module "jenkins_2" {
   log_bucket_name                   = module.jenkins_elb_log_bucket.bucket_id
 }
 
+module "jenkins_3" {
+  source                            = "../../modules/jenkins/jenkins"
+  name                              = "jenkins3"
+  aws_account_and_jenkins_login_ips = var.aws_account_and_jenkins_login_ips
+  jenkins_public_key_name           = aws_key_pair.jenkins.key_name # Or ${var.ssh_key_name}
+  jenkins_instance_profile          = aws_iam_instance_profile.jenkins.name
+  jenkins_wildcard_elb_cert_arn     = aws_acm_certificate.jenkins_wildcard_elb_certificate.arn
+  ami_id                            = "ami-03caf24deed650e2c"
+  instance_type                     = "t3.large"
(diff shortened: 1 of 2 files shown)


terraform/accounts/main/jenkins.tf
```diff
@@ -22,6 +22,20 @@ module "jenkins_2" {
   log_bucket_name                   = module.jenkins_elb_log_bucket.bucket_id
 }
 
+module "jenkins_3" {
+  source                            = "../../modules/jenkins/jenkins"
+  name                              = "jenkins3"
+  aws_account_and_jenkins_login_ips = var.aws_account_and_jenkins_login_ips
+  jenkins_public_key_name           = aws_key_pair.jenkins.key_name # Or ${var.ssh_key_name}
```
bjgill, line 29: `# Or ${var.ssh_key_name}` - what does this mean?

terraform/modules/jenkins/jenkins/ec2.tf
```diff
@@ -9,10 +9,14 @@ resource "aws_instance" "jenkins" {
   iam_instance_profile    = var.jenkins_instance_profile
   key_name                = var.jenkins_public_key_name
   vpc_security_group_ids  = [aws_security_group.jenkins_instance_security_group.id]
+
```
bjgill, line 12: Why this new blank line?

bjgill (approved):
Good catch for the root volume.

Just a couple of minor things - happy for you to merge once addressed.

terraform/accounts/main/jenkins.tf
gidsg: @bjgill this is a copy-paste from the current instance block, but appears to be a reference that you may wish to use your own variable (as defined in https://github.com/alphagov/digitalmarketplace-aws/blob/889d2e369bbd25c939f69d8bec39aa6c4143938d/terraform/accounts/main/jenkins.tf#L2-L3)

terraform/modules/jenkins/jenkins/ec2.tf
gidsg: removed in 889d2e3.

bjgill (approved):

PR #970, opened 2021-08-02
Make new instance primary

gidsg:

Make new instance primary, rename old instance to allow for switching back in case of issues


alex9smith (approved):
👍 ♻️

PR #972, opened 2021-08-05
remove old jenkins instance

gidsg:

We have retired this in favour of jenkins3.


bjgill (approved):
:tada:

PR #982, opened 2021-08-18
Start using invoke for Python-based tasks

bjgill:
So that we can manage this repo in the same way we do all our others.

See https://github.com/alphagov/digitalmarketplace-scripts/pull/649 for an example of how we've done this for other repos. There's a lot of non-Python stuff in the Makefile, so I've left that as-is for the moment in the interests of time.

As part of this, I needed to switch us over to using pip-compile. This has no immediate effect, but might make future dependency management easier.


.github/workflows/test.yml
```diff
@@ -21,6 +21,9 @@ jobs:
           path: venv
           key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements*.txt') }}
 
+      - name: Install developer tools
+        run: make bootstrap
+
       - name: Install dependencies
         run: make requirements-dev
```
phantas, line 28:
```suggestion
        run: invoke requirements-dev
```
likewise for line 31 (I can't add as a suggestion...) which should be replaced for
```
        run: invoke test
```

phantas (changes requested):
couple of changes to github flows, otherwise looks ok 👍

bjgill:
Don't know why that's not working - it's doing the same thing as our other repos. I'll investigate...

bjgill:
OK. That fixes the problem. Once merged, I'll deploy an app to make sure I haven't broken anything.

phantas (approved):
LGTM! 🚀

PR #983, opened 2021-08-18
Point API to new restored database in Staging

phantas:
The last time staging was updated with clean live-like data, the job failed and left the database in an inconsistent state, so we had to do a restore (see https://alphagov.github.io/digitalmarketplace-manual/infrastructure/database-backups.html?highlight=database%20restore#restoring-from-a-backup)

Trello: https://trello.com/c/VmiVQKkA

### DIFF
--- a/vars/staging.yml
+++ b/vars/staging.yml
@@ -5,6 +5,10 @@ instances: 2
 
 api:
   memory: 2GB
+  services:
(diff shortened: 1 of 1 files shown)


bjgill (approved):
Looks correct :rocket:

PR #985, opened 2021-08-19
Revert "Start using invoke for Python-based tasks"

bjgill:
Reverts alphagov/digitalmarketplace-aws#982

It broke the build:

```
+ make preview deploy-app
cf push --no-start --no-route -f <(make -s -C /data/jenkins/jobs/release-app-paas/workspace generate-manifest) -o digitalmarketplace/antivirus-api:release-fd5f3dd --docker-username dmppuller
Using docker repository password from environment variable CF_DOCKER_PASSWORD.
Pushing from manifest to org digitalmarketplace / space preview as suppliers@digitalmarketplace.service.gov.uk...
Using manifest file /dev/fd/63
yaml: control characters are not allowed
FAILED
make: *** [Makefile:114: deploy-app] Error 1
```


daniele-occhipinti (approved):
LGTM.
I haven't tested and I am not the best person to review this. However, given the criticality and that this will be tested straightaway "in the field", I feel I can approve it.

bjgill:
Thanks. I agree that this is quite irregular, and would not normally be OK. However, given that it's a revert to fix a broken build, I think it should be low-risk.

PR #998, opened 2021-09-08
Add ccs techops engineers

ghill95:
- Add .idea to git ignore file
- Add Jack, John, Will and Crispin to AWS


alex9smith (approved):
LGTM!

PR #1005, opened 2021-09-24
Get Kibana variables from tfvars file

bjgill:
https://trello.com/c/cJzG0VYS/132-2-update-log-streaming-lambda-to-send-logs-to-ccss-logit-stacks

It's no longer in JSON format (https://github.com/Crown-Commercial-Service/digitalmarketplace-credentials/pull/370), so we need to change how we extract them.

### DIFF
--- a/kibana/Makefile
+++ b/kibana/Makefile
@@ -6,19 +6,19 @@ requirements:
 
 .PHONY: elasticsearch-url
 elasticsearch-url:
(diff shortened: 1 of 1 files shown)


bjgill:
Yes, this is very fragile. I think that's OK because this script is run very infrequently.

bjgill:
Doesn't work with the CCS Kibana, unfortunately. Not sure why. I can do it manually, though, so it's not worth spending any more time on this

PR #1064, opened 2022-01-20
Correct name of ACM validation DNS entry

bjgill:
Previously, the ACM DNS entry for assets was using the same name as for www. I think this was a copy-paste error, because there exists an assets_acm_name variable that was defined but not used.

I think this should put a stop to the current dueling between assets and www, where terraform flips back and forth because they both have the same name.


tim-s-ccs (approved):

bjgill:
Ah - there was a second thing that needed fixing. Curiously, only for Preview - Staging and Production look fine.

tim-s-ccs (approved):

bjgill:
Deployed successfully in preview. Looks like Preview is now stable.

I'll merge this and deploy to staging and prod tomorrow.

